PR #26, opened 2019-09-10
Fix/sidebar anchor links

bloqhead:
Documented bug within VuePress - https://github.com/vuejs/vuepress/issues/1499

Hopefully this fixes it but it requires more testing.


subnetmarco:
This is not working, the anchors are not redirecting me in the right spot of the documentation.

subnetmarco:
To replicate:
* Go in homepage
* Click Documentation on the header
* Click on "DPs and Data Model'

PR #36, opened 2019-09-15
Fix/sidebar anchor links

bloqhead:


bloqhead:
@subnetmarco I've put together a different approach for caching the documentation page images. I've tested it in multiple major browsers and it works. Safari seems to be the exception and I'm not sure why. I'm using a [method](https://blog.lovemily.me/preload-images-with-pure-javascript/) that has been around for a long time. If this doesn't do the trick, I'll have to instead open an issue with the VuePress team and see if they can shine some light on it.

https://deploy-preview-36--kuma.netlify.com/

In the meantime, I have a fix on the live site that works 100%, but is not ideal because of the manual work required on my part. I would prefer to have it automated, which is what I've done with this new method.

I can explain why it happens. The short version is that it's caused by images above the anchor not being completely loaded yet. Once they load and render their height into the browser, content is pushed down and the "shift" occurs. The anchor navigation when going to a page like "DPs and Data Model" is happening faster than the initial image loading.

PR #37, opened 2019-09-16
Add architecture info in installation pages

joakimr-axis:
Currently, all the downloadable tarballs referred to in the installation
documentation are amd64/x86_64 ones. This patch adds info about that in
the page titles in order for us who run e.g. Ubuntu/aarch64 and similar
Linux + non-PC HW systems to be mislead.

Change-Id: I9b59b1c8b20f6f66119e325650973ff02850f623
Signed-off-by: Joakim Roubert <joakimr@axis.com>

### DIFF
--- a/docs/docs/0.1.2/installation/centos.md
+++ b/docs/docs/0.1.2/installation/centos.md
@@ -1,4 +1,4 @@
-# CentOS
+# CentOS (x86\_64)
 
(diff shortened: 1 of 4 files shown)


subnetmarco:
Thanks for the contribution @joakimr-axis

PR #43, opened 2019-09-18
docs(*) kuma-dp start arguments

jakubdyszkiewicz:
It's more natural and elegant to run kuma-dp with arg vars.


jakubdyszkiewicz:
Back then, the 0.1.2 was the newest version. Now this PR is essentially a duplicate of https://github.com/Kong/kuma-website/pull/61 so I'm closing it in favour to that one.

PR #58, opened 2019-10-25
Script for cutting releases

bloqhead:
Okay, after tons of testing, I think this is ready for review. Here's the rundown:

## What the script does
1. It will clone the `draft` directory to a new directory that is named by the version specified
2. After cloning the directory, it will find and replace all instances of `%%VER%%` in the markdown files with the new version
3. It will add the new sidebar navigation structure for the new version to the `sidebar-nav.js` file (it bases the nav structure on the previous versions, so it may require manual editing on your part if there are new pages under your new release)
4. It will append the new version to the `releases.json` file, which is what the website uses to handle the versioning on the documentation pages, install, etc.

After these steps run, you can run the site locally with `yarn docs:dev` and you should see your new version running accordingly!

## How do I use the script?
You'll want to re-run `npm|yarn install` first and foremost. Once that's done, run `npm link`. This will link the `kumacut` command to your bin so you can run it globally.

### Here's the command breakdown

```
kumacut --help
Usage: kumacut [options] [command]

Options:
  -v, --version     Output the current version of this script.
  -h, --help        output usage information

Commands:
  latest            display the latest version of Kuma
  bump              this will simply cut a new patch and bump the patch number up by 1
  new <type> [ver]  options: major, minor, custom <version>, or it defaults to patch
```

If you want to simply put out a new patch and increment by 1, you can run `kumacut bump`. For anything else, you can run `kumacut new [major|minor|custom]`. For `custom`, you can supply a version number in case you don't want to use the default options.

If you want to simply see what the latest version in the docs is for reference, you can run `kumacut latest`.

Let me know if this is clear, could be better, etc.


package.json
```diff
@@ -1,12 +1,16 @@
 {
-  "name": "konvoy-website",
+  "name": "kuma-website",
   "description": "The website and docs for Konvoy.",
```
yskopets: Let's replace `Konvoy` in the `description`

bloqhead: @jakubdyszkiewicz and I talked about having a directory for the most recent features and changes that isn't yet linked to publicly on the website. So it would be the starting point where changes are made, and then it's cut to a new release with this script.

But we could instead copy the latest and work from that. That would be an easy enough change. @jakubdyszkiewicz what are your thoughts?

yskopets: Ah, it wasn't clear to me that I'm expected to edit files inside `draft` folder.

Then, I have another problem. I've made some changes inside `draft` folder and now I need to preview them.

When I start the app using `yarn docs:dev` and open `http://localhost:8080/docs/draft/`, there is no navigation tree on the left.

Can preview `draft` version without creating a release ?

bloqhead: Ah I overlooked the sidebar setup for the draft version. My apologies for being unclear. I'm going to revise the readme as well to convey the overall flow.

bloqhead:
@yskopets I've revised the script so that when you run the site locally, you can view your work as you go via `http://localhost:8080/docs/DRAFT/`. This version will be accessible on the website to those who know the URL, but not present in the navigation anywhere.

So the flow would go:

1. Run `yarn docs:dev`
2. View your work on `http://localhost:8080/docs/DRAFT/`
3. Make your changes and view the site as they're compiled
4. Once your changes are complete, run the `kumacut` script with the parameters you need

One thing I had to change was the RegEx string I was looking for for updating the version throughout the markdown files when running the kumacut script. Using `%%VER%%` was causing a build error when building routes, so I've changed it to `DRAFT` instead. This will also allow links within the DRAFT documentation files to work properly when viewing locally before cutting a new version and clicking around.

jakubdyszkiewicz:
Just letting you know.
I tested the script and it works great, exactly what I had in mind. :)

bloqhead:
@yskopets you should be able to navigate to `http://localhost:8080/install/draft/` locally now.

bloqhead:
> Cool! It works now!

Awesome! @subnetmarco @jakubdyszkiewicz let me know if there's anything this might need. If not, I can merge it in.

PR #67, opened 2019-11-18
docs: document TrafficRoute REST API

yskopets:
### Summary

* document `TrafficRoute`  REST API


yskopets:
@jakubdyszkiewicz I think, it should be fixed in REST API then (so that we could simply copy-and-paste response into docs)

PR #74, opened 2019-12-02
docs: expand on traffic route docs

devadvocado:


docs/docs/0.3.0/policies/README.md
```diff
@@ -206,6 +209,14 @@ spec:
         service: redis
         version: '2.0'
 ```
+You also need to add labels to your services and use that label as a matching tag. Above, you  see that we utilize the `version` tag to help with canary deployment. Another common use-case is to add a `env` tag as you separate testing, staging, and production environments' services.
```
yskopets: Let's only use term `tag` (and don't use `label`).

yskopets: > You also need to add labels to your services

What is "services" ? Is it a `Dataplane` resource ?

yskopets: Let's move this high-level description of how `TrafficRoute` works above (before the example).

docs/docs/0.3.0/policies/README.md
```diff
@@ -206,6 +209,14 @@ spec:
...
+
+#### Resolving Collision in Traffic Routes
+In the event where two traffic route many have colliding configurations, these are the expected behaviors:
```
yskopets: It worth explaining what does it mean for 2 configurations to collide. Maybe by giving an example

docs/docs/0.3.0/policies/README.md
```diff
@@ -206,6 +209,14 @@ spec:
...
+* Identical Selectors: the policy will be prioritized alphabetically by name
```
yskopets: Let's explain what does "Identical Selectors" mean

yskopets: Let's give an example, e.g. `policy named "A" will win over policy named "B"`

docs/docs/0.3.0/policies/README.md
```diff
@@ -206,6 +209,14 @@ spec:
...
+* More tags: The policy with more matching tags will be prioritized first
```
yskopets: Let's give an example, e.g. `policy that matches "source" dataplanes by 2 tags, e.g. "service" and "env", will win over a policy that matches "source" dataplanes only by 1 tag, e.g. "service"`

docs/docs/0.3.0/policies/README.md
```diff
@@ -206,6 +209,14 @@ spec:
...
+* Specific tags: The policy that match by exact value will be prioritized over ones matched by `'*'`
```
yskopets: Let's give an example, e.g. `service: web` will win over `service: '*'`

docs/docs/0.3.0/policies/README.md
devadvocado: updated and moved up

docs/docs/0.3.0/policies/README.md
devadvocado: expanded on collisions in all three examples with code snippets

docs/docs/0.3.0/policies/README.md
```diff
@@ -209,13 +208,143 @@ spec:
...
-You also need to add labels to your services and use that label as a matching tag. Above, you  see that we utilize the `version` tag to help with canary deployment. Another common use-case is to add a `env` tag as you separate testing, staging, and production environments' services.
 
-#### Resolving Collision in Traffic Routes
-In the event where two traffic route many have colliding configurations, these are the expected behaviors:
-* Identical Selectors: the policy will be prioritized alphabetically by name
-* More tags: The policy with more matching tags will be prioritized first
-* Specific tags: The policy that match by exact value will be prioritized over ones matched by `'*'` 
+The expected outcome is that 1 out of roughly 10 request will be routed to the new version 2. 
+
+### Resolving Collision in Traffic Routes
```
yskopets: Let's move section `Resolving Collision in Traffic Routes` into a separate PR.

The idea is that in `0.3.1` release we will now have 3 resources with identical behaviour of conflict resolution:
* `TrafficRoute`
* `TrafficLog`
* `HealthCheck`

So, it makes sense to change the way we describe it. It's no longer `TrafficRoute`-specific.

docs/docs/0.3.0/policies/README.md
```diff
@@ -209,13 +208,143 @@ spec:
...
+#### Identical Selectors:
+In the two `TrafficRoute` policies below, we observe a collision. Both `route-1` and `route-2` have matching source and destination tags. The only differential is which version of redis they route to.
+```yaml
+apiVersion: kuma.io/v1alpha1
```
yskopets: Let's change the format of these examples from `k8s` to `universal` mode

PR #100, opened 2020-01-14
docs: copy up-to-date overview into DRAFT version

yskopets:
* This PR duplicates #96 but for `DRAFT` version


yskopets:
@subnetmarco We have to update documentation under `/docs/docs/draft` and then backport it into a particular version, if necessary

PR #112, opened 2020-01-24
Normalize font sizes

bloqhead:
Normalized and standardized font sizes and weights on the homepage in various places.


bloqhead:
Screenshots for comparison:

### Before
kuma-font-normalizing-BEFORE

### After
kuma-font-normalizing-AFTER

PR #113, opened 2020-01-27
Docs/networking

subnetmarco:


docs/docs/0.3.2/documentation/README.md
```diff
@@ -1828,7 +1828,13 @@ hostssl all             all             0.0.0.0/0               password
 You can also provide client key and certificate for mTLS using `KUMA_STORE_POSTGRES_TLS_CERT_PATH` and `KUMA_STORE_POSTGRES_TLS_KEY_PATH`.
 This pair can be used for auth-method `cert` described [here](https://www.postgresql.org/docs/9.1/auth-pg-hba-conf.html).
 
-## Ports
+## Networking
+
+Kuma - being an application that wants to improve the underlying connectivity between your services and make the network agnostic to the developers - also comes with some networking requirements itself.
```
yskopets: > network agnostic to the developers

sounds weird. does it mean that network is not tied to particular developers ?

docs/docs/0.3.2/documentation/README.md
```diff
@@ -1828,7 +1828,13 @@ hostssl all             all             0.0.0.0/0               password
...
+
+### kuma-cp ports
+
+First and foremost, the `kuma-cp` application is a server that offers a number of services - some meant for internal consumption by `kuma-dp` data-planes, some meant for external consumption by a client.
```
yskopets: > some meant for external consumption by a client

only 1 client ?
who is the client? a human or a tool ?

docs/docs/0.3.2/documentation/README.md
```diff
@@ -1841,8 +1847,40 @@ When `kuma-cp` starts up, by default it listens on a few ports:
 * `5682`: the HTTP server that provides the Envoy bootstrap configuration when the data-plane starts up.
 * `5683`: the HTTP server that exposes Kuma UI.
 
-## Quickstart
+### kuma-dp ports
+
+When we start a data-plane via `kuma-dp` we expect all the inbound and outbound service traffic to go through it. The inbound and outbound ports are defined in the [dataplane specification](#dataplane-specification) when running in universal mode, while on Kubernetes the service traffic always runs on port `15001`.
```
yskopets: > while on Kubernetes the service traffic always runs on port `15001`.

non-technical users won't understand this part anyway; for technical users it might read weird

docs/docs/0.3.2/documentation/README.md
```diff
@@ -1841,8 +1847,40 @@ When `kuma-cp` starts up, by default it listens on a few ports:
...
+
+In addition to the service traffic ports, the data-plane automatically also opens the `envoy` [administration interface](https://www.envoyproxy.io/docs/envoy/latest/operations/admin) service on the following ports:
```
yskopets: > automatically also opens the `envoy`  administration interface *service*

when you call it a "service", it gives an impression that it should be possible to apply policies to it (which is not the case)

docs/docs/0.3.2/documentation/README.md
```diff
@@ -1841,8 +1847,40 @@ When `kuma-cp` starts up, by default it listens on a few ports:
...
+
+* On Kubernetes, by default on port `9901`.
+* On Universal, by default on the first available port greater or equal than `30001`.
```
yskopets: It's worth to mention the complete interface - `127.0.0.1:9901` and `127.0.0.1:30001` - which makes it clear that it's not accessible from outside of the host

docs/docs/0.3.2/documentation/README.md
```diff
@@ -1841,8 +1847,40 @@ When `kuma-cp` starts up, by default it listens on a few ports:
...
+
+The Envoy administration interface can also be manually configured to listen on any arbitraty port by specifying the `--admin-port` argument when running `kuma-dp`.
```
yskopets: Let's add a link to https://kuma.io/docs/0.3.2/documentation/#envoy

docs/docs/0.3.2/documentation/README.md
```diff
@@ -1841,8 +1847,40 @@ When `kuma-cp` starts up, by default it listens on a few ports:
...
+
+### Service Discovery
+
+Here we are going to be exploring the communication between `kuma-dp` and `kuma-cp`, and the communication between multiple `kuma-dp` to handle our service traffic.
+
+Every time a data-plane (served by `kuma-dp`) registers itself to Kuma, it initiates a gRPC streaming connection to Kuma (served by `kuma-cp`) in order to retrieve the latest configuration and share data to the control-plane.
```
yskopets: > registers itself to Kuma

might be ambiguous. A better phrasing might be

"connects to the Control Plane (`kuma-cp`)"

yskopets: > and share data to the control-plane

is definitely ambiguous.

What data does it share with the Control Plane ? Intercepted traffic ?

docs/docs/0.3.2/documentation/README.md
```diff
@@ -1841,8 +1847,40 @@ When `kuma-cp` starts up, by default it listens on a few ports:
...
+::: tip
+The connection between the data-planes and the control-plane is not on the execution path of the service requests, which means that if the data-plane temporarily loses connection to the control-plane the service traffic won't be affected.
+:::
+
+While doing so, the data-planes also advertise the IP address of each service. The IP address is retrieved:
+
+* On Kubernetes by looking at the address of the `Pod`.
+* On Universal by looking at the inbound listeners that have been configured in the [`inbound` property](#dataplane-specification) of the data-plane specification.
+
+The IP address that's being advertised by every data-plane to the control-plane is also being used to route service traffic from one `kuma-dp` to another `kuma-dp`. This means that Kuma knows at any given time what are all the IP addresses associated to every replica of every service.
```
yskopets: > is **also** being used to

 what are the other usages ? (e.g., you might want to explicitly mention metrics scraping by `Prometheus`)

docs/docs/0.3.2/documentation/README.md
```diff
@@ -1841,8 +1847,40 @@ When `kuma-cp` starts up, by default it listens on a few ports:
...
+
+Because Kuma by design already knows the address to every data-plane - and therefore to every service - it is not required to use Kuma with a third-party service discovery tool or dns resolver, because such tools would not give Kuma any additional information than the one it already knows.
```
yskopets: > because such tools would not give Kuma any additional information than the one it already knows

There is always a use case of proxying traffic to services outside of the mesh. In that case a third-party service discovery tool would enable load-balancing

docs/docs/0.3.2/documentation/README.md
```diff
@@ -1841,8 +1847,40 @@ When `kuma-cp` starts up, by default it listens on a few ports:
...
+
+In order for connectivity to work among the `kuma-dp` instances, Kuma also assumes a flat network topology: this means that every data-plane must be able to consume another data-plane by directly sending requests to its IP address. This is true also in the case of multi-region or multi-datacenter setups.
+
+<center>
+<img src="/images/docs/0.3.2/diagram-16.jpg" alt="" style="padding-top: 20px; padding-bottom: 10px;"/>
+</center>
 
-If you need help, you can chat with the [Community](/community) where you can ask questions, contribute back to Kuma and send feedback.
+As illustrated in the picture above, every `kuma-dp` must be able to consume every other `kuma-dp` on the specific ports that govern service traffic, as described in the `kuma-dp` [ports section](#kuma-dp-ports).
```
yskopets: > **consume** every other `kuma-dp`

"consume" might be ambiguous

subnetmarco:
> Don't forget to apply the same changes to the DRAFT version (otherwise they will not be included into docs for the next release)

Can this be removed as a requirement, and before every release we take the latest version of the documentation and updated it for the latest version number?

Applying changes in both places is very taxing not just for our contributions, but also for external contributors.

yskopets:
> Can this be removed as a requirement, and before every release we take the latest version of the documentation and updated it for the latest version number?

We need to have a way to document features of the upcoming release rather than of the latest one.

The solution for it right now is to use `DRAFT` version.

If it's inconvenient, we might need to revisit the way we manage content on `kuma.io`.

docs/docs/0.3.2/documentation/README.md
```diff
@@ -1841,8 +1847,40 @@ When `kuma-cp` starts up, by default it listens on a few ports:
...
+* On Kubernetes, by default at `127.0.0.1:9901`.
+* On Universal, by default on the first available port greater or equal than `30001`, like `127.0.01:30001`.
+
+The Envoy administration interface can also be [manually configured](#envoy) to listen on any arbitraty port by specifying the `--admin-port` argument when running `kuma-dp`.
+
+### Service Discovery
+
+Here we are going to be exploring the communication between `kuma-dp` and `kuma-cp`, and the communication between multiple `kuma-dp` to handle our service traffic.
+
+Every time a data-plane (served by `kuma-dp`) connects to the control-plane Kuma, it initiates a gRPC streaming connection to Kuma (served by `kuma-cp`) in order to retrieve the latest policiy configuration, and send diagnostic information to the control-plane.
```
jakubdyszkiewicz: > connects to the control-plane Kuma

shouldn't this be?

> connects to the Kuma control-plane

or

> connects to the control-plane

PR #127, opened 2020-02-11
Documentation restructure

bloqhead:
The goal for this change was to break up the first child pages in the sidebar out as standalone pages for SEO purposes (anchor pages are not good SEO when they're top-level pages of importance).

For review, I've revised the structure and sidebar for the `0.3.2` folder _only_. This changes requires a bit of work but could prove more manageable.

## Pros
- Each documentation topic is filed under a parent category as its own markdown file, instead of piled into one singular `README.md` file
  - This makes knowing what and where to make your documentation edits much easier
  - Instead of multiple people hacking away on a single root markdown file in a folder, the work can be spread out which can reduce merge conflicts
- We can have more finite control of each sidebar group's behavior (e.g. if you want one group to be collapsible, you can specify this, but disable it on another group) -- to provide some clarity, [this is a sidebar group](https://share.getcloudapp.com/KouW0nzz) in VuePress ([read more about Sidebar Groups here](https://vuepress.vuejs.org/theme/default-theme-config.html#sidebar-groups))
- This addresses the issue of problematic first-child URLs that have a hash in them
  - For example, this URL: `/docs/0.3.2/documentation/#kumactl` becomes this: `/docs/0.3.2/documentation/kumactl/`
- We can still use markdown headings to generate child navigation items in the sidebar automatically for things further down in the sidebar navigation hierarchy
- This approaches negates the need for modifying core theme functionality of VuePress itself, which would put us into a position where we can't upgrade to future versions of VuePress due to having to maintain function overrides

## Cons
- The [sidebar](https://github.com/Kong/kuma-website/blob/chore/seo-doc-restructure/docs/.vuepress/site-config/sidebar-nav.js) would have to be managed manually, unless we come up with a way to build navigation hierarchy via Node. The lack of more automated management of the VuePress sidebar is actually a point of contention among the VuePress community that still needs a solid approach. There are plugins available that extend its functionality, but I found that they were not a fit for our needs and introduced their own set of caveats
- This change would mean we have to go through and split up the documentation content accordingly for each version prior to `0.3.2`. This can be time consuming but the process is straight forward. It does require some markdown heading revisions in order to make sure the VuePress sidebar is organized properly. The changes I've made to the [`0.3.2`](https://github.com/Kong/kuma-website/tree/chore/seo-doc-restructure/docs/docs/0.3.2) directory are a good starting reference point
  - [Read more about how VuePress leverages headings to generate sidebar links](https://vuepress.vuejs.org/theme/default-theme-config.html#nested-header-links)
- Some links between pages will need to be updated to reflect the new URL changes


bloqhead:
I've revised and restructured ALL of the documentation pages for every version. I've also updated all of the URLs within the doc pages so that they are relative and no longer absolute.

This has many benefits. The sidebar structure is much easier to use and now has collapsible sections (a nice optional baked-in feature of VuePress), and instead of the documentation content being stuck inside a singular `README.md` file in a directory, each page is broken out by its specific content (e.g. `mutual-tls.md`, `enabling-modernization.md`, etc). This will reduce the amount of merge conflicts we face as well.

PR #130, opened 2020-02-21
docs: expand on k8s prereqs

devadvocado:


docs/docs/0.3.2/documentation/README.md
```diff
@@ -71,6 +71,32 @@ When following the installation instructions, `kuma-injector` will be automatica
 <img src="/images/docs/0.2.0/diagram-08.jpg" alt="" style="width: 500px; padding-top: 20px; padding-bottom: 10px;"/>
 </center>
 
+#### Matching Labels in `Deployment` and `Service`
+
+When deploying an application in Kubernetes to run alongside Kuma, you must ensure that for each `Deployment` there is at least one matching `Service`. For example, in [Kuma's demo application](https://github.com/Kong/kuma-demo/blob/master/kubernetes/), the [`Deployment` for the Redis service]((https://github.com/Kong/kuma-demo/blob/master/kubernetes/kuma-demo-aio.yaml#L104))  has the following matchLabels:
```
yskopets: > When deploying an application in Kubernetes to run alongside Kuma

So, user application is secondary ? 😄

yskopets: > you must ensure that for each `Deployment` there is at least one matching `Service`

The more precise statement would be that "every `Pod` must be a part of at least one `Service`".

It's necessary for:
* to give a `Pod` (and its `Dataplane`) the correct identity in the `Mesh` (which is `spiffe://<mesh name>/<service name>`)
* to ensure that Kuma policies can be applied to that `Pod` (and its `Dataplane`). Notice that in Kuma polices we treat `service` tag as mandatory in majority of situations

docs/docs/0.3.2/documentation/README.md
```diff
@@ -71,6 +71,32 @@ When following the installation instructions, `kuma-injector` will be automatica
 <img src="/images/docs/0.2.0/diagram-08.jpg" alt="" style="width: 500px; padding-top: 20px; padding-bottom: 10px;"/>
 </center>
 
+#### Matching Labels in `Deployment` and `Service`
```
yskopets: Sorry, I think that this information is too specific to a particular user from Kuma Slack who made a typo in `Service` or `Deployment`.

I expect that the majority of `k8s` users already understand how a `Service` matches a `Pod`.

So, let's rephrase the text to focus on the part that `Kuma` requires every `Pod` to be a part of at least one `Service`.

It will be shorter and relevant to everyone.

bloqhead: @devadvocado can you assist with the verbiage for this title change?

bloqhead:
This was completely resolved in #152

PR #135, opened 2020-02-26
docs: document how to opt out of sidecar injection into a particular Pod

yskopets:


yskopets:
@subnetmarco No, such behaviour is not supported right now.

If we need one, let's start from the requirements:
* if `kuma.io/sidecar-injection = enabled` annotation is present on a `Pod`, do we still require its `Namespace` to have a `kuma.io/sidecar-injection` label ?
* should we introduce new values for `kuma.io/sidecar-injection` label on a `Namespace`, e.g.
  * `enabled-by-default`
  * `disabled-by-default`
  * `enabled` as an alias for `enabled-by-default`

Personally, I prefer to have an explicit `kuma.io/sidecar-injection` label on a `Namespace`:
* operationally, it helps admins of Kubernetes cluster to clearly understand what exactly `Namespaces` might be affected by `Kuma`
* technically, it helps to limit the scope of `Pods` that have to go through `kuma-injector`

subnetmarco:
> if kuma.io/sidecar-injection = enabled annotation is present on a Pod, do we still require its Namespace to have a kuma.io/sidecar-injection label ?

Basically this would be the behavior:

* If the Namespace has `kuma.io/sidecar-injection = enabled`, by default all the Pods in the namespace will have a sidecar injected, unless `kuma.io/sidecar-injection = disabled` is specified in a Pod. If the annotation `kuma.io/sidecar-injection = enabled` has been specified on both the Namespace and the Pod, basically the behavior will not change and the sidecar will be injected.
* If the Namespace has `kuma.io/sidecar-injection = disabled`, by default no Pods in the namespace will have a sidecar injected, unless `kuma.io/sidecar-injection = enabled` is specified on a Pod. If the annotation `kuma.io/sidecar-injection = disabled` has been specified on both the Namespace and the Pod, basically the behavior will not change and the sidecar will not be injected.

> Personally, I prefer to have an explicit kuma.io/sidecar-injection label on a Namespace:

I think we should still have this, but with the ability to being able to fine-grain the injection at the Pod level too. If the annotation in the Pod and the one in the Namespace are equal, the behavior won't change.

subnetmarco:
> if kuma.io/sidecar-injection = enabled annotation is present on a Pod, do we still require its Namespace to have a kuma.io/sidecar-injection label ?

Basically this would be the behavior:

* If the Namespace has `kuma.io/sidecar-injection = enabled`, by default all the Pods in the namespace will have a sidecar injected, unless `kuma.io/sidecar-injection = disabled` is specified in a Pod. If the annotation `kuma.io/sidecar-injection = enabled` has been specified on both the Namespace and the Pod, basically the behavior will not change and the sidecar will be injected.
* If the Namespace has `kuma.io/sidecar-injection = disabled`, by default no Pods in the namespace will have a sidecar injected, unless `kuma.io/sidecar-injection = enabled` is specified on a Pod. If the annotation `kuma.io/sidecar-injection = disabled` has been specified on both the Namespace and the Pod, basically the behavior will not change and the sidecar will not be injected.

> Personally, I prefer to have an explicit kuma.io/sidecar-injection label on a Namespace:

I think we should still have this, but with the ability to being able to fine-grain the injection at the Pod level too. If the annotation in the Pod and the one in the Namespace are equal, the behavior won't change.

PR #138, opened 2020-02-28
docs(*) Traffic Trace policy

jakubdyszkiewicz:
Recreate TrafficTrace policy PR
https://github.com/Kong/kuma-website/pull/129


docs/.vuepress/site-config/sidebar-nav.js
```diff
@@ -49,7 +49,7 @@ module.exports = {
         "policies/traffic-permissions",
         "policies/traffic-route",
         "policies/traffic-metrics",
-        "policies/traffic-tracing",
```
yskopets: Won't this change break links to `TrafficTrace` policy on all versions prior to `0.3.2` ?

docs/docs/0.3.2/policies/traffic-trace.md
```diff
@@ -0,0 +1,90 @@
+# Traffic Trace
+
+With the `TrafficTrace` policy you can configure tracing on every Kuma DP that belongs to the `Mesh`.
+Note that tracing operates on L7 HTTP traffic, so make sure that selected dataplanes are configured with HTTP Protocol.
+
+You can configure tracing in 3 steps:
+
+1) Configure tracing backend
+
+On Universal:
```
yskopets: I think, it would make sense to have separate lists of steps for `Universal` and `Kubernetes`.

Having them intermixed makes it harder to follow.

jakubdyszkiewicz: But then we need to duplicate the tip and the 3rd step. How can we structure it then?

docs/.vuepress/site-config/sidebar-nav.js
jakubdyszkiewicz: You're right, thanks

docs/docs/0.3.2/policies/traffic-trace.md
yskopets: up to @subnetmarco

jakubdyszkiewicz: I'll merge it for now just to have docs for newest version. We can revise it in a separate PR.

PR #139, opened 2020-02-28
docs: document HTTP support in Kuma

yskopets:


docs/docs/0.3.2/policies/general-notes-about-http-support-in-kuma.md
```diff
@@ -0,0 +1,54 @@
+# General notes about HTTP support in Kuma
```
jakubdyszkiewicz: How about just `HTTP support in Kuma`?

yskopets: or `About HTTP support in Kuma` ?

jakubdyszkiewicz: Ok, I'm fine with either of those.

PR #153, opened 2020-03-12
Changed some logging fields to a number

svenwal:


svenwal:
I have installed it on my ELK stack and discovered that bytes_sent and _received are send as string which limits the usage for sums, average, etc. in the dashboards. Have changed it now (by removing some ")

subnetmarco:
@jakubdyszkiewicz looks good to you?

subnetmarco:
@svenwal I moved the JSON value back as a one-liner because otherwise that breaks the log collection (every new line in the JSON format is also a new line when it's being parsed by the log collector, and therefore will be interpreted as a new log).

PR #162, opened 2020-03-30
docs(policies) log format

jakubdyszkiewicz:


docs/docs/0.4.0/policies/traffic-log.md
```diff
@@ -197,13 +178,13 @@ Internally, `Kuma` [determines traffic protocol](../http-support-in-kuma) based
 The default format string for `TCP` traffic is:
 
 ```
-[%START_TIME%] %KUMA_SOURCE_ADDRESS_WITHOUT_PORT%(%KUMA_SOURCE_SERVICE%)->%UPSTREAM_HOST%(%KUMA_DESTINATION_SERVICE%) took %DURATION%ms, sent %BYTES_SENT% bytes, received: %BYTES_RECEIVED% bytes
+[%START_TIME%] %RESPONSE_FLAGS% %KUMA_MESH% %KUMA_SOURCE_ADDRESS_WITHOUT_PORT%(%KUMA_SOURCE_SERVICE%)->%UPSTREAM_HOST%(%KUMA_DESTINATION_SERVICE%) took %DURATION%ms, sent %BYTES_SENT% bytes, received: %BYTES_RECEIVED% bytes
 ```
 
 The default format string for `HTTP` traffic is:
```
subnetmarco: @jakubdyszkiewicz when specifying the `format` in the logging backend, where do I specify how the `tcp` format should look like and how the `http` format` should look like?

jakubdyszkiewicz: You can only specify one format in logging backend which is then used for both. This is a good point that there should be 2 different fields (tcpFormat, httpFormat?). You can kind of achieve this right now with creating two separate backends and one TrafficLog with `protocol: http` selector and another without it.

Let me create a card to improve it.

subnetmarco: I think this is okay for now, as long as we document it in an "Info" boxes.

PR #168, opened 2020-04-02
docs(kuma-cp) Fault Injection

lobkovilya:
Add documentation for the Fault Injection policy. 
Also, I incremented the version up to  0.4.1, not sure that it is needed.


subnetmarco:
@lobkovilya please do not create a new 0.4.1 version but add the content in the 0.4.0 folder (and open against `next` like you have done). Prior to the release we will take the `0.4.0` in `next` and convert it to `0.4.1`.

docs/docs/0.4.0/documentation/http-api.md
```diff
@@ -1318,6 +1318,164 @@ Example:
...
+  {
+   "match": {
+    "protocol": "http",
+    "service": "backend"
+   }
+  }
+ ],
+ "conf": {
+  "delay": {
+   "percentage": 50,
```
jakubdyszkiewicz, line 1356: Can we pick different value like 50.5 so we show that this is double and not int?

docs/docs/0.4.0/documentation/http-api.md
```diff
@@ -1318,6 +1318,164 @@ Example:
...
+```bash
+curl http://localhost:5681/meshes/default/fault-injections
+```
+```json
+{
+ "items": [
+  {
+   "type": "FaultInjection",
+   "mesh": "default",
+   "name": "fi1.default",
```
jakubdyszkiewicz, line 1432: how about just `fi1`?

docs/docs/0.4.0/policies/fault-injection.md
```diff
@@ -0,0 +1,95 @@
...
+        delay:
+            percentage: 50
+            value: 5s
+        responseBandwidth:
+            limit: 50 mbps
+            percentage: 50 
+```
+
+### Sources & Destinations
+`FaultInjection` is a Connection policy, which means it is applied to the connection between dataplanes. As most of the policies, `FaultInjection` supports the powerful mechanism of matching, which allows you to precisely match source and destination services.
```
jakubdyszkiewicz, line 66: I don't think we use `Connection policy` concept in the docs, so maybe just:
`FaultInjection` is a policy which is applied to the connection between selected dataplanes
?

`source and destination services`
maybe source and destination dataplanes? You can do `service: '*'` and match for other tags

docs/docs/0.4.0/policies/fault-injection.md
```diff
@@ -0,0 +1,95 @@
...
+
+::: warning
+Since `FaultInjection` policy available only for L7 HTTP traffic, `protocol: http` is mandatory tag both for source and destination selector.
```
jakubdyszkiewicz, line 69: The warning at the top tells me the same message. Can we merge those and have only this one?

docs/docs/0.4.0/policies/fault-injection.md
```diff
@@ -0,0 +1,95 @@
...
+
+### HTTP Faults
+
+At least one of the following Faults should be specified.
+#### Abort
+
+Abort defines a configuration of not delivering requests to destination service and replacing the responses from destination dataplane by
+predefined status code.
+
+- httpStatus -  HTTP status code which will be returned to source side
```
jakubdyszkiewicz, line 80: httpStatus -> `httpStatus` and same with other fields

jakubdyszkiewicz:
Is it added in the policies tab? The one with squares and icons? I think we will need an icon for it. Can you add a card for it?

docs/docs/0.4.0/policies/fault-injection.md
```diff
@@ -0,0 +1,92 @@
...
+On Universal:
+
+```yaml
+type: FaultInjection
+mesh: default
+name: fi1
+sources:
+    - match:
+        service: frontend
+        version: "0.1"
```
jakubdyszkiewicz, line 15: We've got required http tag, right? We should add it to examples, otherwise they are not valid.

lobkovilya: Sure, thank you! I missed a couple of places

docs/docs/0.4.0/documentation/http-api.md
```diff
@@ -1339,7 +1339,7 @@ curl http://localhost:5681/meshes/default/fault-injections/fi1
    "match": {
     "protocol": "http",
     "service": "frontend",
-    "version": "0.1"
+    "version": "0.1",
```
jakubdyszkiewicz, line 1342: I don't think this is a valid JSON, also check other places here

lobkovilya: Yep, thanks! A bit of Go deformation

subnetmarco:
@bloqhead to add the logo.

bloqhead:
The Policies page is updated for when this is ready to be merged in.

jakubdyszkiewicz:
merging

PR #176, opened 2020-04-23
Community call form confirmation updates

bloqhead:
Added the agenda link and calendar invite (ICS file) to the Community Call form confirmation.

### To test locally
1. run `yarn docs:dev`
2. go to `http://localhost:8080/community/?form_success=true` (`?form_success=true` is the URL Salesforce will send the user back to after the form is submitted)

[Screenshot](https://share.getcloudapp.com/OAubbeQP)

You should see each link now in the confirmation the user is presented with after they've submitted the form.

### DIFF
--- a/docs/.vuepress/public/community-call-invite.ics
+++ b/docs/.vuepress/public/community-call-invite.ics
@@ -1,0 +1,60 @@
+BEGIN:VCALENDAR
+PRODID:-//Google Inc//Google Calendar 70.9054//EN
+VERSION:2.0
+CALSCALE:GREGORIAN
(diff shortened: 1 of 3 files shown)


subnetmarco:
@bloqhead is this ready with the last updates?

bloqhead:
> @bloqhead is this ready with the last updates?

@subnetmarco it's finished but I need a working Google Calendar link. The one I was given didn't work because the event is not in a public calendar. I'm following up on it Monday.

bloqhead:
@subnetmarco working GCal invite is in place.

PR #189, opened 2020-05-05
fix(installer): gzip check and amazon linux support

subnetmarco:


subnetmarco:
This has been tested and it works.

PR #196, opened 2020-05-13
docs: updating secrets and mtls guides

subnetmarco:


subnetmarco:
@jakubdyszkiewicz I want to make sure that the instructions are correct.

docs/docs/0.5.0/policies/mutual-tls.md
```diff
@@ -1,167 +1,229 @@
...
+```
+:::
+::: tab "kubectl"
 
-   If you are just Getting Started with Kuma, run `kumactl manage ca provided` commands on the same machine where `kuma-cp` is running.
-   :::
+The following command can be executed only on Kubernetes:
+
+```sh
+$ kubectl get secrets -n kuma-system
```
jakubdyszkiewicz, line 116: We need to filter by type `system.kuma.io/secret` otherwise it will show other secrets that are for CP like TLS certs for admission server etc.

docs/docs/0.5.0/policies/mutual-tls.md
```diff
@@ -1,167 +1,229 @@
...
-      # there is no enabledBackend property which means mTLS is enabled
-      backends:
-      - name: ca-1
-        type: builtin # use Builtin CA (a unique Root CA certificate will be generated automatically)
-   " | kubectl apply -f -
-   ```
+Unlike the `builtin` backend, with `provided` we need to first upload the certificate and key that Kuma will use as [Secret resources](/docs/0.5.0/documentation/secrets), and then reference the Secrets to the mTLS configuration.
 
-3. A user can now change CA type to `provided` and **enable** mTLS back after that
+We are responsible for providing the CA root certificate + key and also to manage their lifecycle. Kuma will then use our CA root certificate + Key to automatically provision (and rotate) data plane certificates for every replica of every service.
```
jakubdyszkiewicz, line 130: We or You? Both forms are used in this page. We may suggest us - Kuma developers/maintainers.

docs/docs/0.5.0/policies/mutual-tls.md
jakubdyszkiewicz: kubectl get secrets -n kuma-system --field-selector='type=system.kuma.io/secret'

PR #198, opened 2020-05-13
Fix the v0.5.0 docs

ryysud:
I fixed examples of `Mesh` resource and an example of the command to generate traffic on [the v0.5.0 docs](https://kuma.io/docs/0.5.0/). 

Refs:

- https://kuma.io/docs/0.5.0/policies/mutual-tls/
- https://kuma.io/docs/0.5.0/policies/traffic-metrics/


jakubdyszkiewicz:
Awesome, thank you for the fixes! 🎉 

It was there in #181 but must have been lost during merging.

PR #204, opened 2020-05-23
docs: updated proxy template

subnetmarco:


subnetmarco:
The old documentation was a little bit hard to understand, I tried to rewrite it but can you please verify that this rewrite is indeed correct @jakubdyszkiewicz @nickolaev @lobkovilya @devadvocado

PR #234, opened 2020-07-09
cleaned up Health Check examples

arctiqjacob:
Cleaned up Health Check example:
```
The HealthCheck "proxy-to-random-sports-team" is invalid: 
* spec.conf.interval: must have a positive value
* spec.conf.timeout: must have a positive value
* spec.conf.unhealthyThreshold: must have a positive value
* spec.conf.healthyThreshold: must have a positive value
```

- removed double reference of kind
- fixed conf key in HealthCheck type


nickolaev:
Fixes https://github.com/kumahq/kuma/issues/893

PR #238, opened 2020-07-16
proxy template modifications

jakubdyszkiewicz:


nickolaev:
Will it make sense to not use `test`, like in `test-cluster` and `x-header: test`? Maybe replace it with `backend` or something that sounds more "production" ready.

jakubdyszkiewicz:
If I'm matching `backend` dataplane there is probably `backend` cluster already 🤔  any ideas for different name?

nickolaev:
`development`, `staging`, `production`?

jakubdyszkiewicz:
environment does not make sense IMHO. I can't find better names. Merging it now to have docs live, we can improve names in separate PR.

PR #251, opened 2020-07-29
fix: Broken link to Envoy docs on command operators

cwurm:
Fixes a broken link to the Envoy docs on command operators across several docs versions.


jakubdyszkiewicz:
Thanks, can you sign the commit?

PR #273, opened 2020-09-07
DP token improvements

jakubdyszkiewicz:


docs/docs/0.7.1/documentation/security.md
```diff
@@ -63,19 +55,39 @@ $ kuma-dp run \
...
+As you can see in the example above, you can generate a token by passing name, mesh, and tags. Control Plane will verify Dataplane against data available in the Token. This means you can generate a token by specifying:
+* only `mesh`. This way you can reuse the token for all dataplanes in a given mesh.
+* `mesh` + `tag` (ex. `kuma.io/service`). This way you can use one token across all instances of given service.
+  Keep in mind that you have to specify all the values. If you have a Dataplane with 2 inbounds, one with `kuma.io/service: backend` and one with `kuma.io/service: backend-admin`, you need to specify both values (`--tag kuma.io/service=backend,backend-admin`).
+* `mesh` + `name` + `tag` (ex. `kuma.io/service`). This way you can use one token for one instance of given service.
+
+
+#### Turn off authentication
+
+You can turn off authentication for testing proposes. Set `KUMA_ADMIN_SERVER_APIS_DATAPLANE_TOKEN_ENABLED` to `false`.
```
nickolaev, line 71: I don't think this is a good practice and we should not advise for it.

docs/docs/0.7.1/documentation/security.md
```diff
@@ -63,19 +55,39 @@ $ kuma-dp run \
   --dataplane-token-file=/tmp/kuma-dp-echo-1-token
 ```
 
-#### Accessing Admin Server from a different machine
+You can also pass Dataplane Token as `KUMA_DATAPLANE_RUNTIME_TOKEN` Environment Variable. 
+
+#### Dataplane Token boundary
+
+As you can see in the example above, you can generate a token by passing name, mesh, and tags. Control Plane will verify Dataplane against data available in the Token. This means you can generate a token by specifying:
+* only `mesh`. This way you can reuse the token for all dataplanes in a given mesh.
```
nickolaev, line 63: Do we have a practical use-case where one would want `tag` only tokens? Or maybe that is not a good practice and we should not advise for it?

nickolaev (approved):

docs/docs/0.7.1/documentation/security.md
jakubdyszkiewicz: There is `for testing purposes` and in addition to that there is a warning in the next line that says it is not for production setup. How else can we say it to users that it should be used only for testing?

docs/docs/0.7.1/documentation/security.md
jakubdyszkiewicz: I don't think placing dp token on only a tag is useful.

docs/docs/0.7.1/documentation/security.md
nickolaev: Well we do have other developer/testing related configurations and we do not advertise them in the docs. Why this one is special and needed here?

jakubdyszkiewicz: What other developer/testing configurations that can be useful for users are we not presenting in the docs?

PR #274, opened 2020-09-07
new flow of the dp

jakubdyszkiewicz:


docs/docs/0.7.1/documentation/dps-and-data-model.md
```diff
@@ -1,28 +1,25 @@
...
 
-When Kuma (`kuma-cp`) runs, it will be waiting for the data-planes to connect and register themselves. In order for a data-plane to successfully run, two things have to happen before being executed:
-
-* There must exist at least one [`Mesh`](../../policies/mesh) in Kuma. By default the system auto-generates a `default` Mesh when the control-plane is run for the first time.
-* There must exist a [`Dataplane`](#dataplane-entity) entity in Kuma **before** the actual data-plane tries to connect to it via `kuma-dp`.
+When Kuma (`kuma-cp`) runs, it will be waiting for the data-planes to connect and register themselves. In order for a data-plane to successfully run, there must exist at least one [`Mesh`](../../policies/mesh) in Kuma. By default the system auto-generates a `default` Mesh when the control-plane is run for the first time.
 
 <center>
-<img src="/images/docs/0.5.0/diagram-10.jpg" alt="" style="width: 500px; padding-top: 20px; padding-bottom: 10px;"/>
+<img src="/images/docs/0.1.0/diagram-10.jpg" alt="" style="width: 500px; padding-top: 20px; padding-bottom: 10px;"/>
```
nickolaev, line 6: hmm is that reference to the older docs right?

Oh actually 0.5.0 does not exist.

docs/docs/0.7.1/documentation/dps-and-data-model.md
```diff
@@ -34,10 +31,7 @@ When we start a new data-plane in Kuma, **two things** have to happen:
 1. The data-plane needs to advertise what service it is responsible for. This is what the `Dataplane` entity does.
 2. The data-plane process needs to start accepting incoming and outgoing requests.
 
-These steps are being executed in **two separate** commands:
-
-1. We register the `Dataplane` object via the `kumactl` or HTTP API.
-2. Once we have registered the DP, we can start it by running `kuma-dp run`.
+To do this, we have to create a file with a `Dataplane` definition and pass it to `kuma-dp run`. This way, data-plane will be registered in the Control Plane and Envoy will start accepting requests.
```
nickolaev, line 34: Is this the recommended way, or the only possible way. I assume the old kumactl apply is still valid for the DataPlane resource. Still this is the preferred way, therefore this is what we recommend here.

docs/docs/0.7.1/documentation/dps-and-data-model.md
```diff
@@ -89,17 +84,22 @@ networking:
...
+  --dataplane-file=dp.yaml \
+  --dataplane-var name=`hostname -s` \
+  --dataplane-var address=192.168.0.2 \
   --dataplane-token-file=/tmp/kuma-dp-backend-1-token
 ```
 
 In order for the `backend` service to successfully consume `redis`, we specify an `outbound` networking section in the `Dataplane` configuration instructing the DP to listen on a new port `10000` and to proxy any outgoing request on port `10000` to the `redis` service. For this to work, we must update our application to consume `redis` on `127.0.0.1:10000`.
 
+::: tip
+You can parametrize your `Dataplane` definition, so you can reuse the same file for many `kuma-dp` instances or even services.
```
nickolaev, line 100: Shall we give examples, where one parametrizes port(s), tag(s) etc? So that users can see how flexible this approach is and adapt the workflow for their needs.

nickolaev (approved):

docs/docs/0.7.1/documentation/dps-and-data-model.md
jakubdyszkiewicz: I gave an example above that parametrizes name and mesh. Should I also parametrize port and tag?

docs/docs/0.7.1/documentation/dps-and-data-model.md
jakubdyszkiewicz: yeah, it exists only in 0.1.0

docs/docs/0.7.1/documentation/dps-and-data-model.md
jakubdyszkiewicz: Old way is still valid, but a new one is recommended. I did not want to confuse users even more about it here. I think the old flow is preferable for integrations like Kubernetes when we convert DP from something.

docs/docs/0.7.1/documentation/dps-and-data-model.md
nickolaev: I think - yes. Just to give users a hint of how far they can go with the parameters.

PR #285, opened 2020-10-02
docs(documentation) transparent proxying section

jakubdyszkiewicz:

### DIFF
--- a/docs/docs/0.7.1/documentation/networking.md
+++ b/docs/docs/0.7.1/documentation/networking.md
@@ -188,3 +188,79 @@ Kuma DNS allocates a VIP for every Service withing a mesh. Then, it creates outb
      }
     },
 ```
+
+## Transparent Proxying
+
+There are two ways of how the service can interact with its sidecar to connect to other services.
+One is explicitly defining outbounds in the Dataplane:
+```yaml
+type: Dataplane
+...
+networking:
+  ...
+  outbound:
+  - port: 10000
+    tags:
+      kuma.io/service: backend
+```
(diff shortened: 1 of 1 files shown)


docs/docs/0.7.1/documentation/networking.md
```diff
@@ -188,3 +188,79 @@ Kuma DNS allocates a VIP for every Service withing a mesh. Then, it creates outb
...
+type: Dataplane
+...
+networking:
+  ...
+  outbound:
+  - port: 10000
+    tags:
+      kuma.io/service: backend
+```
+This approach has the disadvantage that you need to reconfigure the service to use `http://localhost:10000` when it wants to connect with service `backend`.
```
nickolaev, line 206: Explain that this is what we use in Universal.

docs/docs/0.7.1/documentation/networking.md
```diff
@@ -188,3 +188,79 @@ Kuma DNS allocates a VIP for every Service withing a mesh. Then, it creates outb
...
+
+The alternative approach is Transparent Proxying. With Transparent Proxying before we start a service, we apply [`iptables`](https://linux.die.net/man/8/iptables) that intercept all the traffic on VM/Pod and redirect it to Envoy.
```
nickolaev, line 208: Do we have to rpovide some setting to DP to know it is in transparent proxy mode?

jakubdyszkiewicz: nope

docs/docs/0.7.1/documentation/networking.md
jakubdyszkiewicz: added

PR #292, opened 2020-10-12
docs(*) external services

nickolaev:


docs/docs/0.7.2/policies/external-services.md
```diff
@@ -0,0 +1,54 @@
...
+  name: httpbin
+spec:
+  tags:
+    kuma.io/service: httpbin
+    kuma.io/protocol: http
+  networking:
+    address: httpbin.org:80
+    tls:
+      enabled: false
+```
```
jakubdyszkiewicz, line 26: Ok, I apply this on Kubernetes, how can I consume this from my application?

docs/docs/0.7.2/policies/external-services.md
```diff
@@ -0,0 +1,54 @@
...
+  address: httpbin.org:80
+  tls:
+    enabled: false
+```
+
+We will apply the configuration with `kumactl apply -f [..]` or via the [HTTP API](/docs/0.7.2/documentation/http-api).
+:::
+:::: 
+
+ * `tags` the external service can include an arbitrary number of tags, where `kuma.io/service` is a mandatory one. The special `kuma.io/protocol` tag is also taken into account and supports the standard Kuma protocol values. It designates the specific protocol being used by that service.
```
jakubdyszkiewicz, line 49: Can the `kuma.io/service` value be the same as the service within the mesh? If not, state this in the docs.

docs/docs/0.7.2/policies/external-services.md
```diff
@@ -0,0 +1,54 @@
+# External Service
+
+This policy enables consuming services that are not part of the mesh from services running inside it. The default passthrough cluster will allow for accessing any non-mesh host by its domain name or IP address, however as the name of the feature suggests, this pattern leaves the mesh administrator with no tools to aply any policies for such traffic. Therefore, ExternalService resource allows for declaring the desired external resource as a named service within the mesh and enabling the observability, security and traffic manipulation similar to any other service in the mesh.
```
devadvocado, line 3: Would rewrite this paragraph:
- "...name or IP address, however as the name of the feature suggests, this pattern leaves the mesh administrator with no tools to aply any policies for such traffic." Would split the sentence before the "however" and leave out the "as the name suggests" since it is a bit wordy
- Confusing since it states that i have no tools to _apply_ any policies for such traffic but then you mention that you can enable observability, security, and traffic manipulation (also unclear what manipulation you are talking about)

subnetmarco:
@nickolaev please address comments.

PR #294, opened 2020-10-13
docs(security) ingress token

jakubdyszkiewicz:


nickolaev:
Do we need docs about `--type=ingress`?

PR #353, opened 2021-01-21
Install page newsletter form

bloqhead:
This change adds the newsletter form to the bottom of the Install page.

**Note:**  When the form submits, it will redirect back to the Kuma homepage and display the success message there. This is a Pardot form handler limitation we've been facing since day one (SalesForce does not allow AJAX submissions at all on any of their form handlers).

### DIFF
--- a/docs/.vuepress/theme/components/custom/Install.vue
+++ b/docs/.vuepress/theme/components/custom/Install.vue
@@ -1,5 +1,7 @@
 <template>
   <div class="page-container page-container--install">
+    
+    <Content />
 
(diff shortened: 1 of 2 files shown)


bloqhead:
@subnetmarco let me know what you think. [Here's a screenshot](https://share.getcloudapp.com/YEuQ5Qrv). If it looks good, it's ready to merge. The UX and behavior is exactly the same otherwise.

subnetmarco:
Perfect. Merged. @bloqhead when the email is entered, will this redirect back to the installation page?

PR #361, opened 2021-02-01
Add MS3 case study to the homepage

bloqhead:
Also addressed some Tailwind future configuration settings.


subnetmarco:
@bloqhead the case study is in the wrong place, needs to be higher up.

bloqhead:
I am working on some revisions to the homepage that will affect this. So for the time being, I've converted this to a draft until those things are finalized.

PR #372, opened 2021-02-16
docs(kuma-cp) timeout policy

lobkovilya:


docs/docs/1.0.8/policies/timeout.md
```diff
@@ -0,0 +1,70 @@
+# Timeout
+
+This policy enables Kuma to set timeouts on the outbound connections depending on the protocol.
+
+## Usage
+
+As usual, we can apply `sources` and `destinations` selectors to determine what proxy we want to configure (`sources` selector)
+and what outbounds connection of the proxy will be used for this policy (`destinations` selector).
+
+The policy let you configure timeouts for `HTTP`, `GRPC` and `TCP` protocols.
```
Bradamant3, line 10: s/let/lets

comma after `GRPC`

docs/docs/1.0.8/policies/timeout.md
```diff
@@ -0,0 +1,70 @@
+# Timeout
+
+This policy enables Kuma to set timeouts on the outbound connections depending on the protocol.
+
+## Usage
+
+As usual, we can apply `sources` and `destinations` selectors to determine what proxy we want to configure (`sources` selector)
```
Bradamant3, line 7: (OK, starting to copyedit and show y'all how to conform to new/emerging style guide). Rewrite lines 7-8:

"Specify the proxy to configure with the `sources` selector, and the outbound connections from the proxy with the `destinations` selector."

(Fewer words, direct address, clearer instruction)

BUT ... this applies to Universal only, right? I (think I) understand that this is because you have to be explicit about everything with Universal, whereas Kubernetes handles more things under the hood. But it's still awkward here. Leave for now, but we have this issue elsewhere, too. I'll ponder better ways to deal with it throughout.

Bradamant3 (changes requested):
Grammatical change (required) at line 10 of timeout.md, so adding copyedits as well. Thanks!

docs/docs/1.0.8/policies/timeout.md
lobkovilya: Thanks! Fixed

docs/docs/1.0.8/policies/timeout.md
lobkovilya: > "Specify the proxy to configure with the sources selector, and the outbound connections from the proxy with the destinations selector."

I like it, it's indeed more neat and structured. Fixed. 

> BUT ... this applies to Universal only, right? I (think I) understand that this is because you have to be explicit about everything with Universal, whereas Kubernetes handles more things under the hood. But it's still awkward here. Leave for now, but we have this issue elsewhere, too. I'll ponder better ways to deal with it throughout.

No, Timeout policy applies both to Universal and Kubernetes. Do you want to explicitly highlight the fact it works for both?

Bradamant3: What confused me is the explicit assignment of `sources` and `destinations` in the Universal example, but I didn't see them in the K8s example. I assume this is bc K8s takes care of so much automatically? And I'm not sure we need to call it out explicitly. Can we assume users understand the difference? Approving, because seems like we can ship as is and discuss whether we need more detail after release. Thanks!

Bradamant3 (approved):

docs/docs/1.0.8/policies/timeout.md
lobkovilya: Oh, I see what you mean, I'm sorry, I missed `source` and `destinations` in the Kubernetes example, they have to be similar. Thanks for noticing it! Let me fix it

PR #375, opened 2021-02-18
docs(policies) global-scoped secrets

jakubdyszkiewicz:
also added missing HTTP API


docs/docs/1.0.8/security/secrets.md
```diff
@@ -86,13 +88,81 @@ $ echo "value" | base64
 ```
 :::
 
+### Access to the Secret HTTP API
+
+This API requires authentication. Consult [Accessing Admin Server from a different machine](security/#user-to-control-plane-communication) how to configure remote access.
```
Bradamant3, line 93: link URL needs updating to `/docs/1.0.8/security/certificates/#user-to-control-plane-communication` (yeah, my reorg, sorry about that)

Bradamant3 (changes requested):
URL change is the only content thing. But `package-lock.json` has crept back in -- should we add this to `.gitignore`?

jakubdyszkiewicz:
Fixed links, thanks!

Reverted `package-lock.json`. We cannot ignore this file. It is essential for consistent builds
https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json

I run `yarn install` which updated various packages, but that's probably should be done in a separate PR.

Bradamant3 (approved):
LGTM

PR #379, opened 2021-02-23
feat(api) add http api docs for /versions endpoint

bartsmykla:


docs/docs/1.0.8/documentation/http-api.md
```diff
@@ -190,6 +191,52 @@ curl http://localhost:5681/config
 }
 ```
 
+## Supported Envoy versions
+
+### Get supported by Kuma DP versions of Envoy
```
Bradamant3, line 196: I understand what you're saying here, but I think "List supported Envoy versions" would be clearer. We don't need to include "Kuma DP" because the docs context provides that information. And "get" outside the endpoint call is a little confusing.

bartsmykla: Thank you for a comment! :-) Let me go with your suggestion!

bartsmykla: done! :-)

Bradamant3 (approved):

PR #382, opened 2021-02-25
docs(security) improve docs with examples

jakubdyszkiewicz:
* New `control plane to control plane` security section
* More examples-driven docs.
* Provide examples of how to generate certs with kumactl
* K8S kumactl and HELM as separate tabs to have better structure (shorter page etc.). The cost is some duplication between tabs
* Try to keep docs in 1, 2, 3 steps.


Bradamant3 (approved):
LGTM

PR #384, opened 2021-02-26
docs(*) ingress annotation

nickolaev:


docs/docs/1.0.8/documentation/dps-and-data-model.md
```diff
@@ -294,11 +294,7 @@ spec:
...
-
-:::warning
-Failure to apply the `ingress.kubernetes.io/service-upstream=true` annotation will prevent Kuma from taking over the load balancing of the requests and applying policies, therefore it is a required annotation when exposing a Kuma service via a `gateway` data plane proxy.
-:::
+The optimal gateway in Kubernetes mode would be Kong. You can use [Kong for Kubernetes](https://github.com/Kong/kubernetes-ingress-controller) to implement authentication, transformations, and other functionalities across Kubernetes clusters with zero downtime. Using [Kong for Kubernetes](https://github.com/Kong/kubernetes-ingress-controller) with Kuma requires an annotation on every `Service` that you want to pass traffic to [`ingress.kubernetes.io/service-upstream=true`](https://github.com/Kong/kubernetes-ingress-controller/blob/master/docs/references/annotations.md#ingresskubernetesioservice-upstream). This is automatically injected by Kuma for every Kubernetes service that is in a namespace part of the mesh i.e. has `kuma.io/sidecar-injection: enabled`
```
Bradamant3, line 297: This explanation took me a while to parse -- initially it seems to read as though you add the annotation manually, which makes removing the annotation field from the example yaml seem not right. I had to read carefully several times before I understood what the last sentence was saying (about automatically injecting the annotation). Let's go ahead and merge this, and I'll work on simplifying the explanation in a separate PR (need to go away and think about it some).

I also wonder whether we should be less pushy about Kong gateway, since Kuma is not only OSS but CNCF. Yes we want to explain it, but wdyt softening this tone also? I can include a revision in the PR to clarify the annotation explanation.

Bradamant3 (approved):
LGTM, with followup PR next week to clarify some things.

nickolaev:
I moved it to 'next' as this is where it belongs actually.

PR #385, opened 2021-02-26
remove telus testimonial for now

Bradamant3:
### Summary
Remove Telus testimonial.

Not sure about `yarn.lock` -- upgraded locally, didn't expect changes in repo. lmk if I should revert.

PTAL @bloqhead

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -28,11 +28,6 @@ tabs:
 
 # testimonials
 testimonials:
-  - content: 'Kuma reduces complexity and accelerates service reliability with an Envoy-based Service Mesh.'
-    author: 'Luca Maraschi'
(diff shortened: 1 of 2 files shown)


nickolaev (approved):

bloqhead:
@Bradamant3 it looks like the yarn.lock updates are version bumps on a small handful of Babel things. Not sure what triggered it on your end, but as long as everything works, that's totally fine. I can pull the site down and test a build locally with netlify-cli just to double check.

Bradamant3:
@bloqhead I ran `yarn upgrade` to get the site to build locally, so I'm guessing that's what triggered it. But ... I'd have expected my local `yarn.lock` to be in sync with the upstream site if I needed to upgrade. 🤷‍♀️

PR #392, opened 2021-03-05
docs(kuma-cp) info about workdir

lobkovilya:
I'm really struggling to find a suitable place for this piece of information, so I just created a separate file to discuss it here.  @Bradamant3 what could be the best place for this info?


Bradamant3:
Would it make sense to put this in https://kuma.io/docs/1.0.8/security/certificates/? I need to rewrite that whole page to focus on what the user needs to do (too much conceptual explanation up front), but I think for now we could do something like:

```
# PKI certificates

This section addresses the certificate generation in Kuma across the following use-cases:

1. Security between our services, via the [mTLS policy](/docs/1.0.8/policies/mutual-tls/).
2. Security between the Kuma control plane and its data plane proxies, via the data plane proxy token.
3. Security when accessing the control plane.
4. Security when external system access the control plane.

Kuma stores autogenerated certificates and other directories in a working directory. The default value for this directory is `$HOME/.kuma`. You can change the working directory with `KUMA_GENERAL_WORK_DIR`.

:::tip
This section is not to be confused with the [mTLS policy](/docs/1.0.8/policies/mutual-tls/) that we can apply to a [Mesh](/docs/1.0.8/policies/mesh/) to secure service-to-service traffic.
:::
```

Bradamant3 (approved):
LGTM

PR #393, opened 2021-03-05
Remove 0.x versions

bloqhead:
This change removes all 0.x doc files so that the site build will be much lighter. It also resolves build errors in Netlify that occurred due to memory allocation exhaustion, which was occurring because of the amount of files being processed.

I've also done some cleanup to resolve peer dependency issues, and changed a small VuePress config setting to make pages load faster in the browser (see [shouldPrefetch](https://v1.vuepress.vuejs.org/config/#shouldprefetch) in the VuePress docs).


Bradamant3 (approved):
LGTM

PR #411, opened 2021-04-12
Reintroduced the max_old_space flag for the build commands

bloqhead:
VuePress is not optimized well and does not handle garbage collection well at all. this change should remove the build error that occurs part way through the HTML rendering step, which is the final build step.

Signed-off-by: Daryn St. Pierre <daryn.st.pierre@konghq.com>


bloqhead:
@Bradamant3 side note: there seems to be 2 markdown files for the 1.1.2 release. I'm seeing this in the build process:

```
4:26:01 PM: warning Override existing page /blog/2021/kuma-1-1-2/.
```

I also see the 2 files. The date seems to be one day off but the content differs:
* `2021-4-08-kuma-1-1-2.md`
* `2021-4-09-kuma-1-1-2.md`

bloqhead:
I'm merging this in regardless so that we can get the site deployed. VuePress will give the `2021-4-09-kuma-1-1-2.md
` file priority since it's a later date and a newer file, but we should remove the duplicate.

PR #413, opened 2021-04-14
docs(*) ipv6 support

nickolaev:
Signed-off-by: Nikolay Nikolaev <nikolay.nikolaev@konghq.com>

### DIFF
--- a/docs/docs/1.1.2/networking/ipv6.md
+++ b/docs/docs/1.1.2/networking/ipv6.md
@@ -6,10 +6,8 @@ scenarios are continuously verified - standalone Kubernetes and Universal, multi
 
 For the most part any IPv6 setup will work out of the box, but there are some specifics that need to be taken into account:
  
- * when the dataplane proxies are run in IPv6 only environment (i.e. no IPv4), the DNS shall be set to generate relevant
-   IPv6 addresses. Using the standard 240.0.0.0/4 will render the system not able to communicate in the `.mesh` domain. The
-   setup can be done by setting `KUMA_DNS_SERVER_CIDR` to a valid IPv6 CIDR. Please verify there is no overal with a pre-existing
-   network in your environment. An example would be `KUMA_DNS_SERVER_CIDR=fd00:fd00::/64`
+ * when the dataplane proxies are run in IPv6 only environment (i.e. no IPv4 address), the DNS shall be set to generate relevant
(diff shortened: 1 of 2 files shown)


docs/docs/1.1.2/networking/ipv6.md
```diff
@@ -0,0 +1,31 @@
...
+All Kuma entitites do support running in mixed IPv4 and IPv6 environments as well as pure IPv6 setup. This includes
+Global and Remote control planes, the Data plane proxy, the accompanying iptables scripts and the CNI. All major deployment
+scenarios are continuously verified - standalone Kubernetes and Universal, multizone and hybrid.
+
+For the most part any IPv6 setup will work out of the box, but there are some specifics that need to be taken into account:
+ 
+ * when the dataplane proxies are run in IPv6 only environment (i.e. no IPv4), the DNS shall be set to generate relevant
+   IPv6 addresses. Using the standard 240.0.0.0/4 will render the system not able to communicate in the `.mesh` domain. The
+   setup can be done by setting `KUMA_DNS_SERVER_CIDR` to a valid IPv6 CIDR. Please verify there is no overal with a pre-existing
+   network in your environment. An example would be `KUMA_DNS_SERVER_CIDR=fd00:fd00::/64`
```
Bradamant3, line 12: This could be clearer. What do you mean by 'overal'? The way the paragraph is written, it's not clear whether the example is meant to be valid, or invalid. (I assume anyone who wants to use IPv6 would understand, but it's confusing at best. Also, isn't 240.0.0.0/4 IPv4? So would it make more sense to call this out as a separate listitem?

Here's a pattern I'd use for rewriting (with clarifying details added bc I'm not sure what they should be):

* In an IPv6-only environment, set the DNS to generate the IPv6 addresses your environment requires. Set `KUMA_DNS_SERVER_CIDR` to a valid IPv6 CIDR, such as <example>.
* In an environment that includes IPv4 addresses, do not specify the 240.0.0.0/4 range. The system cannot communicate over these addresses in the `.mesh` domain.

lmk if my questions aren't clear!

Bradamant3 (changes requested):
See comment at lines 11-12 in the ipv6.md file. Hope it helps!

Bradamant3 (approved):
LGTM and thanks for clarification!

PR #417, opened 2021-04-20
docs(*) transparen proxy upgrade clarifications

nickolaev:
Signed-off-by: Nikolay Nikolaev <nikolay.nikolaev@konghq.com>


Bradamant3 (approved):
LGTM. Should we also update https://github.com/kumahq/kuma/blob/1.1.4/UPGRADE.md? (looks like it hasn't been touched in forever ... maybe we should put upgrading instructions actually in a separate page in the docs?)

nickolaev:
> LGTM. Should we also update https://github.com/kumahq/kuma/blob/1.1.4/UPGRADE.md? (looks like it hasn't been touched in forever ... maybe we should put upgrading instructions actually in a separate page in the docs?)

My first idea was to put this in the UPGRADE.md, but then realised this is recommended for each and every upgrade on Universal so probably be to have it in all docs.

PR #420, opened 2021-04-27
Capitalize the H

gjtorikian:


Bradamant3:
Thanks so much @gjtorikian -- you've made me more aware of how we miscapitalize GH elsewhere in the Kong docs, too. We'd love more docs/website PRs if you're so inclined, but please to amend commits for DCO (`--signoff` flag which I realize you must know but adding it here for convenience). (I set the DCO check to pass for this PR bc, well, we know you're good to go :D )

PR #427, opened 2021-05-06
feat(*) docs for healthcheck changes

bartsmykla:
- Updated HealthCheck policy docs - combined http and tcp
  manifests in one, as with the last changes you can specify
  both in one manifest, also added `reuseConnection` to examples
- Updated API reference


Bradamant3 (approved):
LGTM

PR #430, opened 2021-05-10
fix spelling mistakes

tharun208:
Used `https://github.com/client9/misspell` to fix spelling mistakes.

Planning to come with a pr to introduce a stage in ci to avoid these mistakes. 

cc @nickolaev @jakubdyszkiewicz


Bradamant3 (approved):
LGTM. Thank you! If you're interested in a programmatic approach, we're planning to implement [vale](https://github.com/errata-ai/vale) for this repo and the Kong OSS docs. It should be a local implementation, however. CI/CD integration isn't really viable for OSS projects -- warnings should be local, and errors -- well, we shouldn't be breaking builds for them. We don't have the resources to fix in timely fashion, and often need to ship quickly, fix errors as fast follows. Drop me a line if you're interesting in helping, though!

PR #437, opened 2021-05-14
fix(*) fix the links for 1.1.6 release

jakubdyszkiewicz:
I think the links were not replaced when the release was done


nickolaev (approved):
@Bradamant3 FYI

subnetmarco:
Have these been regenerated manually or with the replace script?

PR #440, opened 2021-05-17
rewrite policy matching

jakubdyszkiewicz:
Docs on matching were stale.
TrafficPermission is also matched by the most specific one, it was "aggregate" back in the early days.

I added a description of the types of policy (mirroring some concepts from the codebase).
I added a note to every policy which kind of policy it is so the user has more knowledge of how it matches and where it is applied.

### DIFF
--- a/docs/docs/1.1.6/policies/health-check.md
+++ b/docs/docs/1.1.6/policies/health-check.md
@@ -147,4 +147,4 @@ HTTP health checks are executed using HTTP 2
 ## Matching
 
 `HealthCheck` is an [Outbound Connection Policy](how-kuma-chooses-the-right-policy-to-apply.md#outbound-connection-policy).
-You can only use `kuma.io/service` in the `destination` section.
+The only supported value for `destinations.match` is `kuma.io/service`.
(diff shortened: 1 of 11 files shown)


docs/docs/1.1.6/policies/health-check.md
```diff
@@ -143,3 +143,8 @@ HTTP health checks are executed using HTTP 2
   - if **`receive`** section won't be provided or will be empty, checks
     will be performed as "connect only" and will be marked as successful
     when TCP connection will be successfully established.
+
+## Matching
+
+`HealthCheck` is an [Outbound Connection Policy](how-kuma-chooses-the-right-policy-to-apply.md#outbound-connection-policy).
+You can only use `kuma.io/service` in the `destination` section.
```
Bradamant3, line 150: Does this mean that `kuma.io/service` is the only supported value for matching destination? If so, rewrite as

"The only supported value for `destinations.match` is `kuma.io/service`."

(assumes we're ok with the dot notation for nested fields?)

jakubdyszkiewicz: fixed

Bradamant3 (approved):
Thank you!

PR #441, opened 2021-05-17
TrafficPermission with ExternalServices

jakubdyszkiewicz:
This should go to next (it will be released with the next major), but it does not have the newest master


Bradamant3:
@jakubdyszkiewicz I copyedited pretty heavily, plus moved to the right directory. PTAL and merge if you think it's ready, or let me know. Thanks!

Bradamant3:
PTAL @jakubdyszkiewicz -- ready for your final review of edits.

PR #452, opened 2021-06-09
chore(*) rename remote-cp to zone-cp

bartsmykla:
As we are in the process of renaming `Remote CP` to `Zone CP` for the new release, this PR reflects these changes in our GUI


Bradamant3 (approved):
LGTM. I assume we'll have an item about the renaming in the changelog, but can you think of where we should possibly mention it in the actual docs? (I will ponder this question too.)

PR #453, opened 2021-06-09
docs(policies) external service allow renegotiation

jakubdyszkiewicz:

### DIFF
--- a/docs/docs/1.2.0/policies/external-services.md
+++ b/docs/docs/1.2.0/policies/external-services.md
@@ -23,8 +23,15 @@ spec:
     kuma.io/protocol: http
   networking:
     address: httpbin.org:443
-    tls:
+    tls: # optional
       enabled: true
+      allowRenegotiation: false
+      caCert: # one of inline, inlineString, secret
+        inline: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0t... # Base64 encoded cert
+      clientCert: # one of inline, inlineString, secret
+        secret: clientCert
+      clientKey: # one of inline, inlineString, secret
+        secret: clientKey
 ```
 
 Then apply the configuration with `kubectl apply -f [..]`.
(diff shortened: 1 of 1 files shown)


nickolaev (approved):

docs/docs/1.2.0/policies/external-services.md
```diff
@@ -23,8 +23,15 @@ spec:
     kuma.io/protocol: http
   networking:
     address: httpbin.org:443
-    tls:
+    tls: # optional
       enabled: true
+      allowRenegotiation: false
+      caCert: # one of inline, inlineString, secret
+        inline: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0t... # Base64 encoded cert
```
jpeach, line 30: Is this actually required to be base64 encoded? If so, why? YAML has enough string quoting types that base64 doesn't need to be required, and the Datasource proto doesn't document any base64 requirement.

docs/docs/1.2.0/policies/external-services.md
```diff
@@ -68,27 +82,32 @@ networking:
...
-     * `clientCert` the client certificate for mTLS
-     * `clientKey` the client key for mTLS
- 
+* `tags` the external service can include an arbitrary number of tags, where `kuma.io/service` is mandatory. The special `kuma.io/protocol` tag is also taken into account and supports the standard Kuma protocol values. It designates the specific protocol for the service.
+* ` networking` describes the networking configuration of the external service
+    * `address` is the address where the external service can be reached.
+    * `tls` is the section to configure the TLS originator when consuming the external service
+        * `enabled` turns on and off the TLS origination.
+        * `allowRenegotiation` turns on and off TLS renegotiation. It's not recommended enabling this for [security reasons](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/tls.proto).
+          However, some servers require this setting to fetch client certificate after TLS handshake. TLS renegotiation is not available in TLS v1.3.
```
jpeach, line 108:
```suggestion
However, some servers require this setting to fetch client the certificate after the TLS handshake. TLS renegotiation is not available in TLS v1.3.
```

Ignore the lack of leading whitespace in the suggestion ... chrome won't let me paste whitespace in there :shrug:

docs/docs/1.2.0/policies/external-services.md
Bradamant3: I'm wondering whether we want to provide a hard-coded value here, or whether a variable indicator and letting the user know they need to provide their own value for the cert would be better.

Docs team tends to lean toward variable use, in which case it would look like:

```suggestion
        inline: {YOUR_CERT_VALUE}... # Your Base64 encoded cert
```

PR #457, opened 2021-06-15
feat(*) document metrics endpoint filtering

jpeach:


docs/docs/1.2.0/policies/traffic-metrics.md
```diff
@@ -85,6 +85,9 @@ metrics:
 
 Both snippets from above instruct `Kuma` to configure every dataplane in the mesh `default` to expose an HTTP endpoint with `Prometheus` metrics on port `5670` and URI path `/metrics`.
 
+The exposed metrics endpoint forwards to the standard Envoy [Prometheus metrics endpoint](https://www.envoyproxy.io/docs/envoy/latest/operations/admin#get--stats?format=prometheus) and supports the same query parameters.
+By default this endpoint returns all available metrics, but if only a subset of the available metrics are needed, the `filter` query parameter can be used to limit the results to metrics whose names match a given regular expression.
```
Bradamant3, line 89:
```suggestion
The metrics endpoint is forwarded to the standard Envoy [Prometheus metrics endpoint](https://www.envoyproxy.io/docs/envoy/latest/operations/admin#get--stats?format=prometheus) and supports the same query parameters.
You can pass the `filter` query parameter to limit the results to metrics whose names match a given regular expression. By default all available metrics are returned.
```

Bradamant3 (commented):
Approval isn't needed to merge, so making a suggestion only and not requiring it. The whole page (and all the policies pages) will get overhauls soon-ishly, but I figured I'd start showing some cleaner text now :-)

PR #459, opened 2021-06-16
chore(*) modify dns docs after change in defaults

bartsmykla:
As the dataplane proxy build in DNS is enabled by default now
I reorganized the docs a little bit to reflect the change


Bradamant3 (approved):
LGTM as MVP; merging so I can pull down more easily to continue the reorganization work.

PR #460, opened 2021-06-16
Finish reorganization of DNS topic

Bradamant3:
Merged @bartsmykla's MVP PR to next, worked on edits and finishing reorg in new branch. PTAL @bartsmykla or anybody else available to make sure I didn't change technical details in the process.

Note I removed this text:

`This DNS configuration is critical for multi-zone deployments as multi-zone requires .mesh addresses to be used to resolve services across the zones in a service mesh. Workloads that are connected as an ExternalService do not require this .mesh resolution. Take special note of your deployment topology as Universal mode and Kubernetes mode have difference configuration steps, both of which are covered within this page.`

... because it seems important but its placement on the page was confusing, and it probably belongs elsewhere too.

This PR also merges `master` back to `next` and partially updates `next`.

Only one page needs review. Netlify preview: https://deploy-preview-460--kuma.netlify.app/docs/1.2.0/networking/dns/


jpeach (approved):
LGTM

PR #469, opened 2021-06-24
Reword warning on arbitrary tags in destination selectors

scottslowe:
This PR rewords and clarifies the warning regarding the use of arbitrary tags on `destinations` selectors for Kuma policies. It also fixes a typo in the same warning.


jpeach (approved):
LGTM

PR #470, opened 2021-06-24
Fix Timeout Matching section

scottslowe:
This PR fixes an incorrect reference in the "Matching" section for the Timeout policy. This section incorrectly referenced the Retry policy.


jpeach (approved):
LGTM

PR #481, opened 2021-07-05
fix(docs) add a warning for Zone CP in-memory

lobkovilya:

### DIFF
--- a/docs/docs/1.2.1/deployments/multi-zone.md
+++ b/docs/docs/1.2.1/deployments/multi-zone.md
@@ -33,10 +33,10 @@ To implement easy service connectivity, Kuma ships with:
 * **Zone Ingress Proxy**: Kuma provides an out of the box `zone-ingress` proxy mode that will be used to enable traffic to enter a zone from another zone. It can be scaled horizontally. Each zone must have a `zone-ingress` proxy deployed. 
 
 :::tip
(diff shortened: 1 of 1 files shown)


nickolaev (approved):

docs/docs/1.2.1/deployments/multi-zone.md
```diff
@@ -132,6 +132,11 @@ $ KUMA_MODE=zone \
 
 Where `<zone-name>` is the name of the zone matching one of the Zone resources to be created at the Global CP. `<global-zone-sync-address>` is the public address as obtained during the Global CP deployment step.
 
+::: warning
+In the example above Zone CP is running with in-memory type of store. It's not possible to scale Zone CP while using in-memory store.
+Please consider using Postgres.
+:::
+
```
Bradamant3, line 139:
```suggestion
::: warning
In the example above Zone CP is running with in-memory type of store. It's not possible to scale Zone CP while using in-memory store.
Please consider using Postgres.
:::
Note that this example runs the zone control plane with an in-memory datastore. This approach is not recommended for product because it does not scale. Consider running with Postgres instead.
```

Bradamant3 (approved):
oops, pushed fix to mistake I made in suggestion, with apologies!

PR #485, opened 2021-07-19
fix(docs) improve docs for Zone Ingress

lobkovilya:
As it was fairly noted by @gerardborst, docs are not quite clear when it comes to the addresses ZoneIngress deals with. This PR aims to solve it.

Fix: https://github.com/kumahq/kuma/issues/2014


Bradamant3 (approved):
LGTM

Bradamant3:
We need to update the docs for 1.2.2 with the same changes -- @lobkovilya can you add to this PR please?

Bradamant3 (approved):
Thank you!

PR #486, opened 2021-07-22
typo

cpu601:
correct typo


Bradamant3:
@cpu601 hey Christoph, thanks for the docs fix -- can you add the DCO signoff to your commit please? (We really need to document this better, so sorry!)

PR #491, opened 2021-07-29
Add documentation for virtual-outbound

lahabana:


lahabana:
I thought I'd open the doc PR so that we're clear on the way it works and if you see any possible issue :)

Bradamant3:
> I thought I'd open the doc PR so that we're clear on the way it works and if you see any possible issue :)

This is brilliant -- maybe we can also use this one as a pilot for figuring out comments for generated policy refdocs, since we have a bit of time for it. I see mostly tech-writer nits at the moment -- lmk whether you'd prefer raw comments, comments as suggestions, or a commit directly to the PR.

lahabana:
For nits please do them as suggestions (this helps me improve).

PR #500, opened 2021-08-06
Remove namespace creation in helm install doc

lahabana:
The namespace is now created by the helm chart

### DIFF
--- a/docs/docs/1.2.3/installation/helm.md
+++ b/docs/docs/1.2.3/installation/helm.md
@@ -27,7 +27,6 @@ Once the repo is added, all following updates can be fetched with `helm repo upd
 At this point we can install and run Kuma using the following commands. We could use any Kubernetes namespace to install Kuma, by default we suggest using `kuma-system`:
 
 ```sh
-$ kubectl create namespace kuma-system
 $ helm install --namespace kuma-system kuma kuma/kuma
 ```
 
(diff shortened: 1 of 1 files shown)


austince (approved):

austince:
Wait, sorry – where is it created in the Helm chart? All I see is the patch: https://github.com/kumahq/kuma/blob/master/deployments/charts/kuma/templates/pre-install-patch-namespace-job.yaml

docs/docs/1.2.3/installation/helm.md
```diff
@@ -27,7 +27,6 @@ Once the repo is added, all following updates can be fetched with `helm repo upd
 At this point we can install and run Kuma using the following commands. We could use any Kubernetes namespace to install Kuma, by default we suggest using `kuma-system`:
 
 ```sh
-$ kubectl create namespace kuma-system
 $ helm install --namespace kuma-system kuma kuma/kuma
```
austince, line 30:
```suggestion
$ helm install --create-namespace --namespace kuma-system kuma kuma/kuma
```
Maybe this is what we want instead?

lahabana: This seems correct indeed.
Seems like I misunderstood something when we talked about it with @bartsmykla.
@bartsmykla  can you double check that what @austince is saying matches what you said earlier?

Bradamant3:
Added do not merge label until @bartsmykla reviews -- then you can remove and I'll give it some writer love if needed.

docs/docs/1.2.3/installation/helm.md
bartsmykla: it's exactly what I was thinking about

bartsmykla (approved):
lgtm

PR #510, opened 2021-08-23
docs(kuma-cp) permissive mTLS mode

lobkovilya:

### DIFF
--- a/docs/docs/1.2.3/policies/mutual-tls.md
+++ b/docs/docs/1.2.3/policies/mutual-tls.md
@@ -187,6 +187,48 @@ A few considerations:
 * The `dpCert` configuration determines how often Kuma should automatically rotate the certificates assigned to every data plane proxy.
 * The Secrets must exist before referencing them in a `provided` backend.
 
+## Permissive mTLS
+
+Kuma provides a convenient way to migrate existing workloads to the mTLS mesh with zero downtime. In order to do so 
+`PERMISSIVE` mode has to be enabled. 
+
+:::: tabs :options="{ useUrlFragment: false }"
+
+::: tab "Kubernetes"
+```yaml
+apiVersion: kuma.io/v1alpha1
+kind: Mesh
+metadata:
+  name: default
+spec:
+  mtls:
+    enabledBackend: ca-1
+    backends:
+      - name: ca-1
+        type: builtin
+        mode: PERMISSIVE # supported values: STRICT, PERMISSIVE
+```
+:::
+
+::: tab "Universal"
+```yaml
+type: Mesh
+name: default
+mtls:
+  enabledBackend: ca-1
+  backends:
+    - name: ca-1
+      type: builtin
+      mode: PERMISSIVE # supported values: STRICT, PERMISSIVE
+```
+:::
+
+::::
+
+Permissive mTLS mode encrypts outbound connections the same way as strict mTLS mode, but inbound connections on the server-side
+accept both TLS and plaintext. This allows migrating servers to the mTLS mesh while clients are still don't belong to the mesh.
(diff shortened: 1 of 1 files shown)


docs/docs/1.2.3/policies/mutual-tls.md
```diff
@@ -187,6 +187,48 @@ A few considerations:
 * The `dpCert` configuration determines how often Kuma should automatically rotate the certificates assigned to every data plane proxy.
 * The Secrets must exist before referencing them in a `provided` backend.
 
+## Permissive mTLS
+
+Kuma provides a convenient way to migrate existing workloads to the mTLS mesh with zero downtime. In order to do so 
+`PERMISSIVE` mode has to be enabled.
```
Bradamant3, line 193:
```suggestion
In version 1.3.0 and later, Kuma provides `PERMISSIVE` mTLS mode to let you migrate existing workloads with zero downtime:
```

This ... seems like something we should maybe also encourage users not to leave on? Or at the very least warn about insecure incoming connections? Or ... something?

I'm also not sure about the order of things. Might be better to put the explanation at lines 228-230 before the examples also?

docs/docs/1.2.3/policies/mutual-tls.md
```diff
@@ -187,6 +187,48 @@ A few considerations:
...
+      type: builtin
+      mode: PERMISSIVE # supported values: STRICT, PERMISSIVE
+```
+:::
+
+::::
+
+Permissive mTLS mode encrypts outbound connections the same way as strict mTLS mode, but inbound connections on the server-side
+accept both TLS and plaintext. This allows migrating servers to the mTLS mesh while clients are still don't belong to the mesh.
+The case when the client and server already had TLS is supported as well.
```
Bradamant3, line 230:
```suggestion
Permissive mTLS mode encrypts outbound connections the same way as strict mTLS mode, but inbound connections on the server-side
accept both TLS and plaintext. This lets you migrate servers to an mTLS mesh before their clients. It also supports the case where the client and server already implement TLS.
```

Bradamant3 (changes requested):
Some copyedits and a question. I'm also happy to do the changes if you'd rather. Thanks!

docs/docs/1.2.3/policies/mutual-tls.md
```diff
@@ -187,6 +187,48 @@ A few considerations:
...
+
```
jakubdyszkiewicz, line 231: Not a blocker but I'm missing here a guide on how can I migrate an app with existing TLS. I bet we will see this question on Kuma slack.

Maybe it should be a part of a bigger guide that let say we have 3 apps, 1 <-> 2 <-> 3. There is custom TLS between 2 <-> 3 and we present how to gradually introduce them to a mesh with mTLS without any downtime.
Just an idea, it can go to the backlog, but we may take this into account when we rewrite docs.

docs/docs/1.2.3/policies/mutual-tls.md
jakubdyszkiewicz: Good point. I'm missing the information on when can I switch from permissive to strict mTLS. Metrics are in progress?

docs/docs/1.2.3/policies/mutual-tls.md
lobkovilya: I think we can do it as a tutorial rather than theoretical steps?

docs/docs/1.2.3/policies/mutual-tls.md
lobkovilya: Added a warning: 

image

The message and format is not final, I know @Bradamant3 doesn't like a lot of frames :)

PR #512, opened 2021-08-24
docs(http-api) adjust insight fields

jakubdyszkiewicz:
I had to change some fields after initial implementation https://github.com/kumahq/kuma-website/pull/505


Bradamant3:
Skipping approval and merging bc there's a lot going on right before release.

PR #513, opened 2021-08-24
Docs/external ratelimit matching

parkanzky:
RateLimit can now be appleid to an ExternalService, but there are some caveats. RateLimit was previously an exclusively inbound policy. Now, if RateLimit is applied to an ExternalService that policy is applied on the outbound.


parkanzky:
Needs to be against `next`, not master

PR #531, opened 2021-09-09
docs(http-api) adjust zones docs

tomaszwylezek:
### Summary

- We had wrong urls to zones+insights
- Added config into zones+insights subscription (as it was added recently)

Signed-off-by: Tomasz Wylężek <tomwylezek@gmail.com>


Bradamant3 (approved):
LGTM

PR #537, opened 2021-09-14
correct typos

curx:


Bradamant3 (approved):
Should have checked this one before 536 -- partial dupe. No merge conflicts, so approving as is.

PR #543, opened 2021-09-21
Fix a spelling error.

jpeach:
Fix the spelling of "Kuma".

Signed-off-by: James Peach <james.peach@konghq.com>


Bradamant3 (approved):
LGTM and thanks and maybe someday we'll get these all fixed up :-(

PR #552, opened 2021-10-07
docs(*) new section "External Services and Locality Aware Load Balancing"

lobkovilya:
* copy `1.3.1` dir to `4.4.4` (is it right workflow?)
* add a new section about external services and locality-aware lb

Link for the review: https://github.com/kumahq/kuma-website/pull/552/files#diff-5152b469f8f50653ac8a5f473b5108278903b48b6560e053780b7f51c5cdae1bR118


Bradamant3:
@lobkovilya yes this is exactly right, thank you so much for getting it done (was on my list for today).

PR #575, opened 2021-11-02
Add Helm chart version to install instructions

### LINKED ISSUE #522: Helm install instructions should specify version
With each version release of Kuma, the Helm install instructions should specify that is the version being installed. With any Helm install documentation version you are always getting the latest version of Kuma.

From:
`$ helm install --create-namespace --namespace kuma-system kuma kuma/kuma`

To:

`$ helm install --version x.y.z --create-namespace --namespace kuma-system kuma kuma/kuma`

The `helm install` command specifies the version of the chart, not the version of Kuma, so that needs to be taken into consideration for each version release.

hhoover:
Before this change, users would install the latest
version of the Helm chart (and Kuma) no matter what
the desired outcome was. By specifying a chart
version to helm install, users will get the exact
version of Kuma they desire based on the version
of docs they are reading. Chart versions are based
on versions available using:
`helm search repo kuma -l`

Reference: https://helm.sh/docs/helm/helm_install
fixes #522


Bradamant3 (approved):
LGTM ❤️

PR #580, opened 2021-11-05
docs(api) add resources to GlobalInsights + more

bartsmykla:
Structure of GlobalInsights changed to allow return more of global
resources

Signed-off-by: Bart Smykla <bartek@smykla.com>


jpeach:
This is an API break; is that intended?

Bradamant3:
@bartsmykla how do we want to address James's question?

bartsmykla:
yes it is, but it's within the same version, as the one before was not released

PR #616, opened 2021-11-29
docs(security) token expiration, revocation, signing key rotation

jakubdyszkiewicz:
Opening against master for now, will change to next once it's ready.

* Extracted DP and Zone Ingress auth to a separate page
* Added info about expiration, revocationa and signing key rotation to DP token
* Added info about expiration, revocationa and signing key rotation to Zone Ingress token

I can bring this back to `certificates.md` but that page is already has a lot of info, sections and subsections


Bradamant3:
retargeted to next and moved files where they needed to be.

Bradamant3 (approved):
Could use a bit of a copyedit, but will do as part of a release prep PR (sorry but being time-efficient here :D )

PR #636, opened 2021-12-23
docs(security): data plane proxy mesh membership

jakubdyszkiewicz:
Mesh membership functionality


Bradamant3:
leaving as is for now in case we get rid of semver before I leave, in which case we can do things differently 🤞

lahabana (approved):

PR #647, opened 2022-01-11
Multi zone guide refresh

jakubdyszkiewicz:
Refresh of https://github.com/kumahq/kuma-website/pull/402


docs/docs/1.4.1/deployments/how-multi-zone-works.md
```diff
@@ -0,0 +1,31 @@
+# About multi-zone deployments
+
+Kuma supports the use case where multiple service meshes run in different zones. This use case includes hybrid deployments where you run your services on Kubernetes and in Universal mode, on VMs or elsewhere. Your mesh environment can include multiple isolated service meshes (multi-tenancy), and workloads running in different regions, on different clouds, or in different datacenters. A zone can be a Kubernetes cluster, a VPC, or any other deployment you need to include in the same distributed mesh environment.
```
lahabana, line 3: > Kuma supports running your service mesh in multiple zones. It is even possible to run with a mix of Kubernetes and Universal zones.
Probably shorter and simpler.

docs/docs/1.4.1/deployments/how-multi-zone-works.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Components of a multi-zone deployment
+
+A multi-zone deployment includes:
+
+* The **global control plane**. Accepts connections only from zone control planes -- not from any data plane proxies. It provides configurations for the global [policies](/policies) that are applied to the data plane proxies.
+* The **zone control planes**. Accept connections from data plane proxies that are started in the same underlying zone, and connect to the global control plane to fetch policies for the data plane proxies that connect to them.
+* The **data plane proxies**. Connect to the zone control plane in the same zone.
+* The **zone ingress** Provides ingress to services in different zones. If a zone consumes services only, and does not provide them, the zone does not require this component.
```
lahabana, line 12: I think it's missing a work in "If a zone consumes services only" Did you mean "If a zone consumers only services local to its zone"?

docs/docs/1.4.1/deployments/how-to-multi-zone.md
```diff
@@ -0,0 +1,311 @@
...
+- Set up cross-zone communication between data plane proxies
+
+## Set up the global control plane
+
+The global control plane must run on a dedicated cluster, and cannot be assigned to a zone.
+
+:::: tabs :options="{ useUrlFragment: false }"
+::: tab "Kubernetes"
+
+The global control plane on Kubernetes must reside on its own Kubernetes cluster, to keep its CRDs separate from the CRDs the remote control planes create during synchronization.
```
lahabana, line 17: Is CRD the right word? To me a CRD is the definition of a resource, instances of CRDs are just resources. So the reason we need them separated is to keep the global resources separated from the zone resources no?

docs/docs/1.4.1/deployments/how-to-multi-zone.md
```diff
@@ -0,0 +1,311 @@
...
+1.  Find the external IP and port of the `global-remote-sync` service in the `kuma-system` namespace:
+
+    ```bash
+    kubectl get services -n kuma-system
+    NAMESPACE     NAME                   TYPE           CLUSTER-IP      EXTERNAL-IP      PORT(S)                                                                  AGE
+    kuma-system   global-remote-sync     LoadBalancer   10.105.9.10     35.226.196.103   5685:30685/TCP                                                           89s
+    kuma-system   kuma-control-plane     ClusterIP      10.105.12.133   <none>           5681/TCP,443/TCP,5676/TCP,5677/TCP,5678/TCP,5679/TCP,5682/TCP,5653/UDP   90s
+    ```
+
+    In this example the value is `35.226.196.103:5685`. You pass this as the value of `<global-kds-address>` when you set up the remote control planes.
```
lahabana, line 34: The 3685 appears here magically. you might want to reference the network page and mention that by default KDS runs on 3685 no?

docs/docs/1.4.1/deployments/how-to-multi-zone.md
```diff
@@ -0,0 +1,311 @@
...
+```
+
+And if your HTTP clients take the standard default port 80, you can the port value and run either of:
+
+```
+<kuma-enabled-pod>$ curl http://echo-server_echo-example_svc_1010.mesh
+<kuma-enabled-pod>$ curl http://echo-server.echo-example.svc.1010.mesh
+```
+
+Because Kuma on Kubernetes relies on transparent proxy, `kuma-dp` listens on port 80 for all virtual IPs that are assigned to services in the `.mesh` DNS zone. The DNS names are rendered RFC compatible by replacing underscores with dots.
```
lahabana, line 213: It might make sense to reference virtual-outbound to do more flexible things with hostnames and ports.

lahabana (commented):
Some comments but it's a great improvement!

docs/docs/1.4.1/deployments/how-to-multi-zone.md
jakubdyszkiewicz: 5685 right? I added reference to ports.

docs/docs/1.4.1/deployments/how-multi-zone-works.md
jakubdyszkiewicz: While I agree it's shorter and simpler and the one above is wordy, those 2 sentences do not explain what the zone is.

docs/docs/1.4.1/deployments/how-multi-zone-works.md
jakubdyszkiewicz: Zone A can consumer services only from Zone A, but Zone B can also consume services from Zone A, so I'm not sure it's the same

Bradamant3: And neither option correctly explains zone ingress, right? (Yes I know I wrote the original. It's wrong.) I think we can trim to

```suggestion
* The **zone ingress** Provides ingress to services in different zones.
```

docs/docs/1.4.1/deployments/how-multi-zone-works.md
Bradamant3: Yeah, some of this belongs in a higher-level "about Kuma" topic, not here. Originally I was trying to trim and associate a bunch o' different bits of explanation from the current live doc. How about something like

```suggestion
Kuma supports the a service mesh that is distributed across multiple zones. Different zones in the same mesh can run on Kubernetes or in Universal mode, in different regions, on different cloud providers, or in different datacenters. The only thing that zones have in common is their connection to the same global control plane.
```

docs/docs/1.4.1/deployments/how-multi-zone-works.md
jakubdyszkiewicz: > Provides ingress to services in different zones

For me, it sounds like Ingress in zone A provides an entry point to services in zone B, C, D, which is not correct.

How about "Provides ingress to local services for different zones"

docs/docs/1.4.1/deployments/how-multi-zone-works.md
lahabana, line 3: I was suggesting only replacing the first 2 sentences not the whole paragraph.
```suggestion
Kuma supports running your service mesh in multiple zones. It is even possible to run with a mix of Kubernetes and Universal zones. Your mesh environment can include multiple isolated service meshes (multi-tenancy), and workloads running in different regions, on different clouds, or in different datacenters. A zone can be a Kubernetes cluster, a VPC, or any other deployment you need to include in the same distributed mesh environment.
```

docs/docs/1.4.1/deployments/how-multi-zone-works.md
lahabana: I like the way you reworded that!

docs/docs/1.4.1/deployments/how-multi-zone-works.md
```diff
@@ -0,0 +1,31 @@
...
+The DNS resolver is embedded in each data plane proxy. It resolves each service address to a virtual IP address for all service-to-service communication.
+
+The global control plane and the zone control planes communicate to synchronize resources such as Kuma policy configurations over Kuma Discovery Service (KDS), which is a protocol based on xDS.
+
+<center>
+<img src="/images/docs/0.6.0/distributed-diagram.jpg" alt="" style="width: 500px; padding-top: 20px; padding-bottom: 10px;"/>
+</center>
+
+:::tip
+A zone ingress is not an API gateway. Instead, it is specific to internal cross-zone communication within the mesh. API gateways are supported in Kuma [gateway mode](/docs/1.1.1/documentation/dps-and-data-model/#gateway) which can be deployed in addition to zone ingresses.
```
lahabana, line 30: Use relative paths so that we stop getting bad versions :)

docs/docs/1.4.1/deployments/how-to-multi-zone.md
```diff
@@ -0,0 +1,312 @@
...
+
+You must [enable mTLS](../policies/mutual-tls.md) for cross-zone communication between services.
+
+Kuma uses the Server Name Indication field, part of the TLS protocol, as a way to pass routing information cross zones. Thus, the mTLS is mandatory to enable cross-zone service communication.
+
+### Zone Ingress requirements
+
+Cross-zone communication between services is available only if Zone Ingress has a public address and public port.
+
+On Kubernetes, Kuma automatically tries to pick up the public address and port. Depending your load balancing implementation, you might need to wait a few minutes for Kuma to get the address.
```
lahabana, line 179:
```suggestion
On Kubernetes, Kuma automatically tries to pick up the public address and port. Depending on your load balancing implementation, you might need to wait a few minutes for Kuma to get the address.
```

lahabana (approved):
LGTM though if I were I'd wait for #649 to be merged to get the link checker as this probably breaks some stuff.

PR #649, opened 2022-01-11
ci: add link checker on preview deployment

### LINKED ISSUE #566: Broken link Checker in CI
Using for example a GitHub action:
- https://github.com/gaurav-nelson/github-action-markdown-link-check
- https://github.com/lycheeverse/lychee-action

lahabana:
- Add action that crawls the netlify preview and looks for broken links
- Fix broken links in all of 1.4.1
- Always use relative paths in 1.4.1 (this makes moving new versions easier).

The job is currently only running for 1.4.1 to avoid having to fix all the links

Fix #566


docs/docs/1.4.1/deployments/multi-zone.md
```diff
@@ -14,8 +14,8 @@ We can support multiple isolated service meshes thanks to Kuma's multi-tenancy s
...
+* **Global**: this control plane will be used to configure the global Service Mesh [policies](../policies) that we want to apply to our data plane proxies. Data plane proxies **cannot** connect directly to a global control plane, but can connect to `zone` control planes that are being deployed on each underlying zone that we want to include as part of the Service Mesh (can be a Kubernetes cluster, or VM based). Only one deployment of the global control plane is required, and it can be scaled horizontally.
+* **Zone**: we are going to have as many zone control planes as the number of underlying Kubernetes or VM zones that we want to include in a Kuma [mesh](../../policies/mesh/). Zone control planes accept connections from data plane proxies that are started in the same underlying zone, and they connect to the `global` control plane to fetch their service mesh policies. Zone control plane policy APIs are read-only and **cannot** accept Service Mesh policies to be directly configured on them. They can be scaled horizontally within their zone.
```
michaelbeaumont: Why the switch from absolute to relative? I feel like we should stick with one, except maybe for sibling pages. Not a blocker at all for this PR though.

I somewhat prefer absolute for readability and maintenance, because it's easier to figure out what the right link is and easier to find links that need changing when moving a page. I guess introducing the link checker makes that irrelevant, though.

michaelbeaumont (approved):

docs/docs/1.4.1/deployments/multi-zone.md
lahabana: I intend to switch everything to relative once I get the CI worked out.
The advantage of relative is that releasing a new version is copy pasting the current "trunk" version and no need to change all inline links.

michaelbeaumont: Yeah, the root of the absolute version should be `docs/docs/1.4.1` (or whatever version) and not `docs` but that may or may not be easy to change.

lahabana: Yeah unfortunately I can't find a way to do so :(

Bradamant3: wrt relative vs absolute -- the docs already contain an odd mix of absolute and relative links. Tossing my departing hat in the ring to support relative links for the same reasons as Charly, and also bc root of absolute links is confusing, etc. But yeah Mike's point about moving pages/changing content at page level is fair.

PR #671, opened 2022-02-07
docs(networking) rewrite transparent proxy and dns

jakubdyszkiewicz:
* Assume across the networking docs that DNS in dp is the default
* Extract DNS in the CP to the separate doc with warning about deprecation
* Remove duplication of setup instructions between DNS and Transparent Proxying page
* Add reachable services

### DIFF
--- a/docs/docs/dev/networking/dns.md
+++ b/docs/docs/dev/networking/dns.md
@@ -1,45 +1,51 @@
 # DNS
 
-As of Kuma version 1.2.0, DNS on the data plane proxy is enabled by default on Kubernetes. You can also continue to deploy with [DNS on the control plane](#control-plane-dns).
+Kuma ships with DNS resolver to provide service naming - a mapping of hostname to Virtual IPs (VIPs) of services registered in Kuma.
 
-## Data plane proxy DNS
+The usage of Kuma DNS is only relevant when [transparent proxying](../transparent-proxying) is used.
 
-In this mode, all name lookups are handled locally by the data plane proxy. This approach allows for more robust handling of name resolution.
+## How it works
 
-On Kubernetes, this is the default. You must enable it manually on universal deployments.
+Kuma DNS server responds to type `A` and `AAAA` DNS requests, and answers with `A` or `AAAAA` records, for example ```redis.mesh. 60 IN A  240.0.0.100``` or ```redis.mesh. 60 IN AAAAA  fd00:fd00::100```.
 
-### Universal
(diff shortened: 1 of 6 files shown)


docs/docs/dev/networking/dns.md
```diff
@@ -223,3 +161,10 @@ Kuma DNS allocates a VIP for every service within a mesh. Then, it creates an ou
      }
     },
 ```
+
+::: tip
+The following setup will work when `serviceVipEnabled=true` which is a default value.
+
+The preferred way to define hostnames is using [Virtual Outbounds](../policies/virtual-outbound).
+Virtual Outbounds also makes it possible to define dynamic hostnames using specific tags or to expose services on a different port.
```
lahabana, line 169: Might be worth mentioning that serviceVipEnabled will be set to false in the future?

docs/docs/dev/networking/transparent-proxying.md
```diff
@@ -155,29 +134,65 @@ spec:
...
-  --env-var KUMA_RUNTIME_KUBERNETES_SIDECAR_TRAFFIC_EXCLUDE_OUTBOUND_PORTS=5678,8900
-```
+Execute `kumactl install transparent-proxy --help` to see available options.
+::::
 
-When deploying Kuma with HELM, use `controlPlane.envVar` value
+### Reachable Services
+
+By default, every data plane proxy in the mesh follows every other data plane proxy.
+This may lead to a performance problems in larger deployments of the mesh.
```
lahabana, line 155: `a performance problem` or `performance problems`

lahabana (approved):

docs/docs/dev/networking/dns.md
```diff
@@ -1,185 +1,123 @@
...
-```shell
-kumactl install control-plane \
-  --env-var KUMA_RUNTIME_KUBERNETES_INJECTOR_BUILTIN_DNS_ENABLED=false
-```
+This mode implements advanced networking techniques, so take special care for the following cases:
 
-:::
-::: tab "Helm"
+ * The mode can safely be used with the [Kuma CNI plugin](cni/).
+ * In mixed IPv4 and IPv6 environments, it's recommended that you specify an [IPv6 virtual IP CIDR](ipv6/).
```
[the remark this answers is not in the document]
jakubdyszkiewicz: I'm also not sure. Let's keep it maybe...

docs/docs/dev/networking/dns.md
jakubdyszkiewicz: I'd rather hold off with this until we are 100% sure we want to get rid of `serviceVipEnabled`.

PR #687, opened 2022-02-15
cleanup(dps): make gateway and zone-ingress dedicated pages

lahabana:
- Reorder things to make it clear that folks using k8s don't rely care about the dp entity
- Move gateway and Zone-Ingress to their own docs to make the dps-and-data-model docs lighter


jakubdyszkiewicz (approved):
love this refactor

PR #688, opened 2022-02-15
docs(kuma-cp): Add summary of health check methods

parkanzky:
Signed-off-by: Paul Parkanzky <paul.parkanzky@konghq.com>

Add summary of health check methods available within Kuma.

### DIFF
--- a/docs/docs/dev/documentation/health.md
+++ b/docs/docs/dev/documentation/health.md
@@ -1,168 +1,32 @@
-# Health checks
+# Dataplane Health
 
-Health is an important aspect of a microservice architecture. Load balancers use the health status
-of a service to select an endpoint. Orchestrators, such as Kubernetes, use service status
-to manage container lifecycles. Also, users want the service state to be observable through the GUI or CLI.
+Health is an important aspect of a microservice architecture. Kuma can use health status
+to select endpoints for communication between dataplane proxies.
+Orchestrators, such as Kubernetes, use service health status to manage container lifecycles.
+Also, users want the service state to be observable through the GUI or CLI.
+
 
 Kuma supports several health checking methods:
 
-* [Health Check](../../policies/health-check) Policy
-
-  An **active** Kuma policy which configures a dataplane proxy to send extra traffic
-  to other dataplane proxies in order to evaluate their health.
-
-* [Circuit Breaker](../../policies/circuit-breaker) Policy
+## [Circuit Breaker](../../policies/circuit-breaker) Policy
 
   A **passive** Kuma policy which configures a dataplane proxy to monitor its existing
-  mesh traffic in order to evaluate dataplane health.
+  mesh traffic in order to evaluate dataplane health. The dataplane can be configured to
+  respond to a widely configurable range of errors and events that it may detect in communication
+  with remote endpoints.
 
-* [Kubernetes](#kubernetes-probes) and [Universal](#universal-probes) Service Probes
+## [Kubernetes](../../policies/service-health-probes/#kubernetes-probes) and [Universal](../../policies/service-health-probes/#universal-probes) Service Probes
 
   Configuration of centralized health probing of services, either directly by Kuma Control Plane,
-  or by the underlying platform, such as Kubernetes.
+  or by the underlying platform, such as Kubernetes.  These can detect problems from the
+  Control Plane's perspective, and propagate failures to the entire mesh. However, it is necessary
+  for the Control Plane to be available, unlike policies which operate independently on the
+  dataplane itself.
 
-Kuma is able to track the status of the Envoy proxy. If grpc stream with Envoy is disconnected then Kuma considers this 
-proxy as offline, but we still send the traffic regardless of that, because this status is designed to track the connection
-between `kuma-cp` and `kuma-dp`. 
+## [Health Check](../../policies/health-check) Policy
 
-Also, every `inbound` in the Dataplane model has `health` section:
-
-```yaml
-type: Dataplane
-mesh: default
-name: web-01
-networking:
-  address: 127.0.0.1
-  inbound:
-    - port: 11011
-      servicePort: 11012
-      health:
-        ready: true
-      tags:
-        kuma.io/service: backend
-        kuma.io/protocol: http
-```
-
-This `health.ready` status is intended to show the status of the service itself. It is set differently depending on 
-the environment ([Kubernetes](#kubernetes-probes) or [Universal](#universal-probes)), but it's treated the same way 
-regardless of the environment:
-
-- if `health.ready` is true or `health` section is missing - Kuma considers the inbound as healthy and includes it 
-  into load balancing set.
-- if `health.ready` is false -  Kuma doesn't include the inbound into load balancing set.
-
-Also, `health.ready` is used to compute the status of the Dataplanes and Service. You can see these statuses both in Kuma GUI and Kuma CLI:
-
-- if proxy status is `Offline`, then Dataplane is `Offline`:
-- if proxy status is `Online`:
-  - if all inbounds are ready then Dataplane is `Online`
-  - if all inbounds are not ready then Dataplane is `Offline`
-  - if at least one of the inbounds is not ready then Dataplane is `Partially degraded` 
-  - if inbound is not ready then it's not included in the load-balancer set which means it doesn't receive the traffic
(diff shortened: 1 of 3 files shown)


docs/docs/dev/documentation/health.md
```diff
@@ -1,13 +1,25 @@
...
 
-Health is an extremely important part of the microservice architecture, load balancers rely on the 
-health status of the service while picking endpoint from load balancer set. Also, users want the service state to be 
-observable through the GUI or CLI. Orchestrators like Kubernetes also want to know the service status to manage 
-lifecycle of the containers. 
-
-Kuma supports several aspects of the health checking. There are two policies which allows configuring active and passive
-health checks:
-[Health Check](../../policies/health-check) and [Circuit Breaker](../../policies/circuit-breaker).
+Health is an important aspect of a microservice architecture. Load balancers use the health status
```
lahabana, line 3: It might worth saying that these health tools will enable us to select endpoints to communicate to between DPPs. That's really what we're talking about. I think using the term "load balancer" here would just confuse users.

docs/docs/dev/documentation/health.md
```diff
@@ -1,13 +1,25 @@
...
+of a service to select an endpoint. Orchestrators, such as Kubernetes, use service status
+to manage container lifecycles. Also, users want the service state to be observable through the GUI or CLI.
+
+Kuma supports several health checking methods:
+
+* [Health Check](../../policies/health-check) Policy
+
+  An **active** Kuma policy which configures a dataplane proxy to send extra traffic
+  to other dataplane proxies in order to evaluate their health.
```
lahabana, line 12: Would it make sense for each of these to put some "advantage/drawbacks" ?

docs/docs/dev/documentation/health.md
```diff
@@ -1,13 +1,25 @@
...
+
+* [Circuit Breaker](../../policies/circuit-breaker) Policy
+
+  A **passive** Kuma policy which configures a dataplane proxy to monitor its existing
+  mesh traffic in order to evaluate dataplane health.
+
+* [Kubernetes](#kubernetes-probes) and [Universal](#universal-probes) Service Probes
```
lahabana, line 19: Feels like this bit of doc is mixing 2 things:

1. Showing the things users can do to manage service failures
2. Explaining how to use probes

Should this page be "dataplane failure management" (there must be a better name) and we'd move the probe description to policies?

lahabana (commented):
I'd like to have your opinion on a refactor.

docs/docs/dev/documentation/health.md
parkanzky: @lahabana Yes. I originally had a comment about this in the summary, but ended up deleting it. At the very least calling this page "Health Check" is confusing because that is the proper name of one of our policies. I agree, we should divide out the probe config, and expand on the summary.

docs/docs/dev/policies/service-health-probes.md
```diff
@@ -0,0 +1,147 @@
...
+
+- if proxy status is `Offline`, then Dataplane is `Offline`:
+- if proxy status is `Online`:
+  - if all inbounds are ready then Dataplane is `Online`
+  - if all inbounds are not ready then Dataplane is `Offline`
+  - if at least one of the inbounds is not ready then Dataplane is `Partially degraded` 
+  - if inbound is not ready then it's not included in the load-balancer set which means it doesn't receive the traffic
+  - if all inbounds which implement the same service are ready then service is `Online`
+  - if all inbounds which implement the same service are not ready then service is `Offline`
+  - if at least one of the inbounds which implement the same service is not ready then service is `Partially degraded`
```
lahabana, line 43: We probably want to describe the different levels of health.

Each dataplane has multiple inbounds each belonging to a specific service. Each inbound is probed independently.
All inbounds in a mesh that have the same `kuma.io/service` are grouped in what is called a `service`

A service has a status which can take 3 values:

- `Online` if all dataplane inbounds with this service have `health: true`
- `Offline` if all dp inbounds with this service have `health: false`
- `PartiallyDegraded` if any inbounds with this service have `health: false`

A specific dataplane will not be present in the list of endpoints sent to dataplanes trying to communicate with this service if the inbound's matching this service health is false.

There's probably better ways to right this but I feel doing:

1. What is inbound health
2. What is a service
3. What does this mean for advertising to other dataplanes using this service

Helps comprehension.

docs/docs/dev/policies/service-health-probes.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Kubernetes probes
+
+Even if Kubernetes probes are disabled, Kuma takes `pod.status.containerStatuses.ready` in order to fill `dataplane.inbound.health` section.
```
lahabana, line 47: This is useful info but might be better if it's moved further down (let's first describe what we do :) )

docs/docs/dev/policies/service-health-probes.md
```diff
@@ -0,0 +1,147 @@
...
+
```
lahabana, line 48: Worth explaining that in K8s Kuma will update Dataplane health according to the status of the containers implementing the service and the status of the sidecar container (Worth pointing to the [k8s doc related to health](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/)).

If sidecar health is false: --> all inbounds health is false
else for each inbounds we find the container that exposes the pod for the service and if the container state is ready then the inbound is healthy.

docs/docs/dev/policies/service-health-probes.md
```diff
@@ -0,0 +1,147 @@
...
+If you specify `httpGet` probe for the Pod, Kuma will generate a special non-MTLs listener and overrides the probe itself in
```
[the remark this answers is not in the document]
lahabana: Might worth mentioning it works both for liveness and readiness probes.

docs/docs/dev/policies/service-health-probes.md
```diff
@@ -0,0 +1,147 @@
...
+
+The same behaviour could be configured using environment variables: 
+
+- `KUMA_RUNTIME_KUBERNETES_VIRTUAL_PROBES_ENABLED=false`
+- `KUMA_RUNTIME_KUBERNETES_VIRTUAL_PROBES_ENABLED=19001`
+
+
+## Universal probes
+
+On Universal there is no single standard for probing the service. For health checking of the service status on
```
lahabana, line 112:
```suggestion
On Universal there is no single standard for probing the service. For health checking the inbound health on
```

docs/docs/dev/policies/service-health-probes.md
```diff
@@ -0,0 +1,147 @@
...
+Universal Kuma is using Envoy's Health Discovery Service (HDS). Envoy does health checks and reports the status back to Kuma Control Plane.
```
lahabana, line 113:
```suggestion
Universal Kuma can use Envoy's Health Discovery Service (HDS). Envoy does health checks and reports the status back to Kuma Control Plane.
```

lahabana: Maybe there should be something like:

In universal you can implement dataplane health in 2 ways:

1. out of bound by having your orchestrator update the dataplane resource using the http api
2. using Envoy HDS

We'll explain how HDS work:

docs/docs/dev/policies/service-health-probes.md
```diff
@@ -0,0 +1,147 @@
+# Service Health Probes
```
lahabana, line 1: This in more `Dataplane Health Probes` or `Inbound Health Probes` than service? Especially as service is very overloaded :)

lahabana (commented):
That's great! I'm sorry this review is very dense. My comments are more about rewriting the service-health-probes doc so if you want to get the health refactor in and do the rewrite in a follow up PR go ahead!
Though I'd love to see the health-probe rewrite happen as we go as this page really needs clarifying :)

docs/docs/dev/policies/service-health-probes.md
lahabana: Actually I believe that if you do `1` you might want to describe `Dataplane.Probes` which enables to passthrough mTLS for health checks (that's what k8s does).

parkanzky:
Added https://github.com/kumahq/kuma-website/issues/705 to track the health probe page updates. Going to merge what we have so far here.

PR #711, opened 2022-02-23
fix(kubernetes-annotations): kuma.io/mesh is an annotation not a label

michaelbeaumont:
## Summary

`kuma.io/mesh` only works as an annotation, bug from https://github.com/kumahq/kuma-website/pull/678


lahabana (approved):
Thx!

PR #721, opened 2022-03-01
fix(docker): correct version in docker install

### LINKED ISSUE #643: Version string is required to use kumactl docker
### What happened?

The docs suggest to use `kumactl`

```
docker run \
    --net="host" \
    docker.io/kumahq/kumactl: kumactl get meshes
```

However, the displayed command will not work since the version tag was ommited and `kumactl` does not seem to have a `latest` tag

```
Error response from daemon: manifest for kumahq/kumactl:latest not found: manifest unknown: manifest unknown
```

In order to sucessfully run the command, we need to run it as 

```
 docker run \
    --net="host" \
    docker.io/kumahq/kumactl:1.4.1 kumactl get meshes
```

This should be updated in the docs

lahabana:
Fix #643

### DIFF
--- a/docs/docs/1.5.x/installation/docker.md
+++ b/docs/docs/1.5.x/installation/docker.md
@@ -73,12 +73,10 @@ To access Kuma you can navigate to [`127.0.0.1:5681`](http://127.0.0.1:5681) to 
 
 You can use the `kumactl` CLI to perform **read and write** operations on Kuma resources. The `kumactl` binary is a client to the Kuma HTTP API. For example:
 
-```sh
-docker run \
-    --net="host" \
-    docker.io/kumahq/kumactl: kumactl get meshes
-# NAME          mTLS      METRICS      LOGGING   TRACING
-# default       off       off          off       off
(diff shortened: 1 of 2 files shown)


docs/docs/1.5.x/installation/docker.md
```diff
@@ -91,7 +89,7 @@ mtls:
   backends:
   - name: ca-1
     type: builtin" | docker run -i --net="host" \
-  docker.io/kumahq/kumactl: kumactl apply -f -
+  docker.io/kumahq/kumactl:<version> kumactl apply -f -
```
michaelbeaumont, line 92: I guess we can't use the template variable for the current version?

michaelbeaumont (approved):

docs/docs/1.5.x/installation/docker.md
lahabana: Yep I've spent an hour trying to figure out how to do this with no success :( It's a limitation of vuepess

PR #369, opened 2021-02-11
docs: add appProtocol for protocol support in k8s

tharun208:
Signed-off-by: Tharun <rajendrantharun@live.com>

PR #731, opened 2022-03-18
docs(policies): appProtocol support by default

jakubdyszkiewicz:
A long time ago @tharun208 introduced support of configuring protocol using `appProtocol` field in Kubernetes instead of this awkward annotation.
He added docs https://github.com/kumahq/kuma-website/pull/369 but it was lost for some reason.
We now support Kubernetes 1.19+ so this should be a default way of configuring the protocol with the old way still available for backward compatibility.


lahabana (approved):

lahabana:
Worth backporting no?

PR #734, opened 2022-03-18
docs(documentation): Gateway API

jakubdyszkiewicz:
First Gateway API docs with basic usage


docs/docs/dev/documentation/gateway-api.md
```diff
@@ -0,0 +1,146 @@
...
+   ```sh
+   kumactl install demo | kubectl apply -f -
+   ```
+
+2. Add GatewayClass and Gateway
+
+   The Gateway resource represents the proxy instance that handles traffic for a set of Gateway API routes, and a GatewayClass describes characteristics shared by all Gateways of a given type.
+   
+   ```sh
+   echo "apiVersion: gateway.networking.k8s.io/v1alpha2
```
lahabana, line 40:
```suggestion
   echo "
```

lahabana (approved):

PR #753, opened 2022-04-01
feat(docs): updated docs for http api

lukidzi:
Added missing options and fixed paths

Signed-off-by: Łukasz Dziedziak <lukidzi@gmail.com>

### DIFF
--- a/docs/docs/dev/documentation/http-api.md
+++ b/docs/docs/dev/documentation/http-api.md
@@ -2956,7 +2956,7 @@ curl http://localhost:5681/zone-ingresses
 
 Request: `GET /zone-ingress/{name}`
 
-Response: `200 OK` with ZoneIngress entity 
+Response: `200 OK` with ZoneIngress entity
 
 Example:
 ```bash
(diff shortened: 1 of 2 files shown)


docs/docs/dev/documentation/http-api.md
```diff
@@ -63,10 +63,14 @@ By default the API Server is listening on port `5681` (HTTP) and on `5682` (HTTP
...
 * `/zones+insights/{name}`
-* [`/zonesegresses`](#list-zone-egresses)
-* [`/zonesegresses/{name}`](#get-zone-egress)
-* [`/zonesegressoverviews`](#list-zone-egress-overviews)
-* [`/zonesegressoverviews/{name}`](#get-zone-egress-overview)
+* [`/zone-ingresses`](#list-zone-ingresses)
+* [`/zone-ingresses/{name}`](#get-zone-ingress)
+* [`/zoneingresses+insights`](#list-zone-ingress-overviews)
+* [`/zoneingresses+insights/{name}`](#get-zone-ingress-overview)
+* [`/zone-egresses`](#list-zone-egresses)
```
jakubdyszkiewicz, line 70: I think the endpoint is `/zoneegresses`, no?
```
❯❯❯ curl localhost:5681/zoneegresses
{
 "total": 0,
 "items": [],
 "next": null
}

❯❯❯ curl localhost:5681/zone-egresses
404: Page Not Found⏎
```

lukidzi: It was connected to a change in Kuma API. We agreed that we won't change it now so I've reverted to the previous.

lahabana (approved):
LGTM but should this be backported to 1.5.x too?

PR #792, opened 2022-05-03
fix: correct broken links

lahabana:


docs/.vuepress/config.js
```diff
@@ -31,6 +31,64 @@ const dirTree = require('directory-tree')
 const path = require("path");
 const fs = require("fs");
 const versions = require("./versions.js");
+const installMethods = [
```
parkanzky, line 34: Is this all related to broken links checker not exiting correctly?

parkanzky (approved):
LGTM

docs/.vuepress/config.js
lahabana: Nope it's just cleanup from a weirdly placed file

.github/workflows/ci.yml
```diff
@@ -22,8 +22,14 @@ jobs:
...
-          max_timeout: 1200
+          max_timeout: 2400
       - name: link checker
+        timeout-minutes: 30
         run: |
           npm install broken-link-checker -g
-          blc $URL -rof --exclude 127.0.0.1 --exclude 192.168 --exclude redis.io  | grep -v '───OK───'
+          if [[ ! `blc $URL/docs/dev -rof --exclude 127.0.0.1 --exclude 192.168 --exclude redis.io > out` ]]; then
+            cat out | grep -v  '───OK───'
+            exit 1
```
michaelbeaumont, line 32: What am I missing, isn't the first line irrelevant if we always `exit 1`? What went wrong with the original line?

lahabana: the `$?` is the exit code of grep not of the cmd so this was always succeeding

parkanzky: You could use $PIPESTATUS[0] to get the exit code of the first command in the pipeline.

parkanzky: Although, looks like $PIPESTATUS is pretty bash-specific. Probably best to keep it compatible with other shells.

michaelbeaumont: Where was `$?` checked? I thought `pipefail` was set with GH actions.

lahabana:
Can't manage to make the job actually fail. I'm going to revert the change in the github action so we at least fix the links

PR #794, opened 2022-05-05
fix(search): add search for the docs

lahabana:
Also fix google indexing by adding meta noindex


michaelbeaumont (approved):
Awesome, this is big.

slonka:
Netlify preview <3

docs/.vuepress/config.js
```diff
@@ -347,8 +352,12 @@ Sitemap: https://kuma.io/sitemap.xml
         let sp = $page.regularPath.split("/");
         if (sp.length > 1) {
           add("docsearch:section", sp[1]);
-          if (sp[1] === "docs" && sp.length > 2) {
-            add("docsearch:docsversion", sp[2]);
+          if ($page.version) {
+            add("docsearch:docsversion", $page.version);
+            if ($page.version !== versions.latestMinor) {
+              // Only index the latest version of the docs
+              add("robots", "noindex follow")
```
slonka, line 359: I think if you're implemeting this, you should remove disallow from `robots.txt` (line [257](https://github.com/kumahq/kuma-website/pull/794/files#diff-52eb86351480925d61947a58caca1ea76d10950ec26533f16f37cb12a3519dffR257)) seen here: https://deploy-preview-794--kuma.netlify.app/robots.txt. Docs on this say:
 
```
Important: For the noindex directive to be effective, the page or resource must not be blocked by a robots.txt file, and it has to be otherwise accessible to the crawler. If the page is blocked by a robots.txt file or the crawler can't access the page, the crawler will never see the noindex directive, and the page can still appear in search results, for example if other pages link to it.
```

https://developers.google.com/search/docs/advanced/crawling/block-indexing

slonka (changes requested):
I think you're both fixing the indexing of old versions of docs and adding the search input. In general I would split these into separate PRs. On the search side 👍 , on the indexing part: 👎  because I think we need to remove the disallow on robots.txt and rely on meta tag.

docs/.vuepress/config.js
lahabana: Good point! Agreed for the PRs I got a little lazy indeed

slonka (approved):

PR #797, opened 2022-05-09
docs(*): fix CB thresholds for Universal example

lobkovilya:


lahabana (approved):
LGTM but why?

lobkovilya:
@lahabana we had `thresholds` in Kubernetes tab but not in Universal tab

PR #799, opened 2022-05-10
fix: make symlink to kumactl absolute, otherwise it does not work

slonka:

Symlinks in the current form don't work, if they are `relative` they are relative in respect to the target, so we are symlinking `/usr/local/bin/kumactl` to `./kumactl` which evaluates to `/usr/local/bin/kumactl` and that's why we get a “loop” of symlinks:

```bash
bash: /usr/local/bin/kumactl: Too many levels of symbolic links
```

We can fix it by doing an absolute link.


bartsmykla (approved):
lgtm

PR #806, opened 2022-05-13
fix: installer script on debian

slonka:

Installer on debian fails with:
```
sh: 21 Bad substitution
sh: 94: Syntax error: redirection unexpected
```

because it uses `bash` specific things when the shell is defined as `sh`. This PR fixes it and also adds a short test file to check if the installer succeeds on the distributions we list in "install" section in the sidebar. 

I also found that:
- in the sidebar "Redhat" was wrongly named "Debian" so I fixed that as well.
- we use an old name for this project in some places - and now that "[konvoy](https://d2iq.com/products/konvoy)" is actually a product we should not use that name 

### DIFF
--- a/docs/.vuepress/public/installer.sh
+++ b/docs/.vuepress/public/installer.sh
@@ -18,7 +18,7 @@
 # download by setting the VERSION environment variable, and you can change
 # the default 64bit architecture by setting the ARCH variable.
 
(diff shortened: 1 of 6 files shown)


michaelbeaumont (approved):
Nice, tests!
(but do we support OSes that don't have bash installed?)

slonka:
Manually checked:

```bash
curl -L https://deploy-preview-806--kuma.netlify.app/installer.sh | bash -
```

So I'm merging it.

Issue #819, opened 2022-05-23
document new application metric proxy


lahabana:
### Description

This new feature needs documentation

PR #827, opened 2022-05-24
docs(*): added applications metrics through kuma docs

lukidzi:
Added description of new application metrics. #819


docs/docs/dev/policies/traffic-metrics.md
```diff
@@ -297,9 +297,14 @@ This feature requires transparent proxy, so it's currently not available for Uni
...
-
-In addition to exposing metrics from the data plane proxies, you might want to expose metrics from applications running next to the proxies.
-
+ 
+In addition to exposing metrics from the data plane proxies, you might want to expose metrics from applications running next to the proxies. You can do this in 2 different ways:
+ 
+* directly from the applications,
+* through Kuma sidecar.
+ 
+### Expose metrics directly from the applications
```
lahabana, line 306: Do we really want to explain this? That has little to do with kuma.

lahabana (commented):
I'll look at this again tomorrow but you should add your new annotation in the annotation reference docs too.

docs/docs/dev/policies/traffic-metrics.md
lukidzi: It's more about configuration, we don't really explain it too deeply. It's the same thing that was explained previously.

docs/docs/dev/policies/traffic-metrics.md
```diff
@@ -332,6 +337,109 @@ Use the Discovery Service of [your choice](https://prometheus.io/docs/prometheus
 
 To consume paths protected by mTLS, you need Traffic Permission that lets Prometheus consume applications.
 
+### Expose metrics through the Kuma sidecar.
+
+Kuma allows scraping Prometheus metrics from the applications endpoint running in the same `Pod` or `VM`. Later those metrics are aggregated and exposed at the same `port/path` as Dataplane metrics. It is possible to configure it at the `Mesh` level, for all the applications in the `Mesh`, or just for specific applications. It's worth mentioning that any configuration change requires redeployment.
```
lahabana, line 342:
```suggestion
Kuma allows scraping Prometheus metrics from the applications endpoint running in the same `Pod` or `VM`. Later those metrics are aggregated and exposed at the same `port/path` as Dataplane metrics. It is possible to configure it at the `Mesh` level, for all the applications in the `Mesh`, or just for specific applications.
This is especially useful when mTLS is enabled and the prometheus scraper doesn't use mTLS.

::: tip
Any configuration change requires redeployment of the dataplane.
:::
```

docs/docs/dev/policies/traffic-metrics.md
```diff
@@ -332,6 +337,109 @@ Use the Discovery Service of [your choice](https://prometheus.io/docs/prometheus
...
+metadata:
+ namespace: kuma-example
+ name: kuma-tcp-echo
+spec:
+ ...
+ template:
+   metadata:
+     ...
+     annotations:
+       prometheus.metrics.kuma.io/aggregate-my-service-enabled: "false"  # causes that configuration from Mesh is disabled and endpoint's metrics won't be exposed
```
lahabana, line 413:
```suggestion
       prometheus.metrics.kuma.io/aggregate-my-service-enabled: "false"  # causes that configuration from Mesh to be disabled and result in this endpoint's metrics to not be exposed
```

docs/docs/dev/policies/traffic-metrics.md
```diff
@@ -332,6 +337,109 @@ Use the Discovery Service of [your choice](https://prometheus.io/docs/prometheus
...
+metrics:
+ type: prometheus
+ conf:
+   path: /metrics/overridden
+   aggregate:
+     my-service: # causes that configuration from Mesh is disabled and endpoint's metrics won't be exposed
+       enabled: false
+     other-sidecar:
+       port: 1234 # override port from Mesh
+     application:
```
lahabana, line 436: We screwed up a little, we should have made this an array, that's the recommended way in CRDs... Array should be:

```yaml
- name: my-service
   enabled: false
   port: 1234
 ```

File: docs/docs/dev/reference/kubernetes-annotations.md [hunk 5]
```diff
@@ -422,3 +422,52 @@ spec:
       containers:
         ...
 ```
+
+### `prometheus.metrics.kuma.io/aggregate-<name>-enabled`
+
+Define if `kuma-dp` should scrape metrics from the application that has been defined in the `Mesh` configuration. Default value: `true`.
```
lahabana, line 428: Can we point to the longer doc?

docs/docs/dev/policies/traffic-metrics.md
lahabana: I think it's just contributing to information overload. Let's remove this first option completely.

docs/docs/dev/policies/traffic-metrics.md
lukidzi: My bad didn't know about recommended way. Should I create an issue that we should in the future change this to the list as a no-breaking change(new field)?

lahabana (approved):
LGTM once the list is updated

lahabana (approved):

PR #839, opened 2022-06-01
fix: move images to the public folder

lahabana:
Some images were repeated in the image folder, this is not needed


bartsmykla (approved):
lgtm

PR #859, opened 2022-06-14
docs(*): added information about token reload

lukidzi:
Docs for dynamic reload of the token in `kuma-dp` without restart need.


docs/docs/dev/security/dp-auth.md
```diff
@@ -185,6 +185,10 @@ If the signing key is compromised, we must rotate it and all the tokens that was
    ::::
    All new connections to the control plane now require tokens signed with the new signing key.
 
+### Token rotation
+
+In case you need to generate a new token for Dataplane or you are using Projected Service Account Token on Kubernetes it is possible to reload it dynamically in `kuma-dp`. To enable this behaviour you need to set `kuma-cp` configuration property `dpServer.auth.useTokenPath` to `true`. When the property is enabled `kuma-dp` should detect that the token file has changed and is going to reload the token and use it when a new connection to `kuma-cp` is established.
```
michaelbeaumont, line 190:
```suggestion
If you need to generate a new token for a `Dataplane` or you are using service acccount token projection on Kubernetes, it is possible to configure dynamic token reloading. To enable this behaviour, set the `kuma-cp` configuration property `dpServer.auth.useTokenPath` to `true`. When this property is enabled, `kuma-dp` detects changes to the token file, reloads the token and uses the new value when establishing a new connection to `kuma-cp`.
```

slonka (approved):

PR #860, opened 2022-06-14
fix(search): update the search index when changing version

### LINKED ISSUE #855: Search results direct to an old version
### What happened?

If I'm on a specific version (e.g. 1.4.x) of the docs, and select a different version in the dropdown (e.g. 1.6.x), any search results (from the search box) I then click on direct me to the original version (e.g. 1.4.x) of that result.

lahabana:
Before when we changed the version during the dropdown the search
would happen on the previous version.

This is done as a hack in Vue but couldn't find another way ;)

Fix #855


lahabana:
It's part of the `this.$page` which is a standard vuepress way for page metadata

PR #866, opened 2022-06-24
ci(.github): add installer.sh tests

michaelbeaumont:
Runs `installer.sh` in docker containers for each listed OS.


lahabana (approved):
LGTM but can we also add with KM maybe? :)

lahabana:
Should this just delete `/tests/installer.sh`

PR #877, opened 2022-06-24
docs(*): improve "secure access across services" page

lobkovilya:
* Delete comments from "shell" command examples
* Make instructions look similar across the document
* add helm command
* delete "Control plane to Postgres communication" section (cc @lahabana)


docs/docs/dev/security/certificates.md
```diff
@@ -321,26 +325,3 @@ Define firewall rules on the global control plane to only accept connections fro
 :::tip
 Third-party extensions, cloud implementations or [commercial offerings](/enterprise) may be extending the authentication support.
 :::
-
-## Control plane to Postgres communication
```
jakubdyszkiewicz, line 325: where did it go?

docs/docs/dev/security/certificates.md
```diff
@@ -28,102 +27,83 @@ Note, this CA is not the same CA for service-to-service communication.
...
+helm install --create-namespace --namespace <namespace> kuma kuma/kuma \
+  --set controlPlane.tls.general.secretName=general-tls-certs \
+  --set controlPlane.tls.general.caBundle=$(cat /tmp/ca.crt | base64)
+```
 
-The data plane proxy Injector in the control plane automatically provides the CA to the Kuma DP sidecar so Kuma DP can confirm the control plane identity.
+The data plane proxy Injector in the control plane automatically provides the CA to the Kuma DP sidecar 
+so Kuma DP can confirm the control plane identity.
 
-When configured using the `controlPlane.tls.general.secretName` and `controlPlane.general.caBundle` values in Helm, Kuma will use the specified CA and certificate to protect data plane proxy to control plane traffic. Additionally, if not configured otherwise using options described below, Kuma will also use the specified CA and certificate to protect user to control plane traffic and control plane to control plane traffic when in multizone configurations.
```
jakubdyszkiewicz, line 97: I think that was added by Scott, because it was not obvious what general cert is.

docs/docs/dev/security/certificates.md
```diff
@@ -28,102 +27,83 @@ Note, this CA is not the same CA for service-to-service communication.
...
 ```sh
-kubectl create secret generic general-tls-certs -n kuma-system \
+kubectl create secret generic general-tls-certs -n <namespace> \
   --from-file=tls.crt=/tmp/tls.crt \
   --from-file=tls.key=/tmp/tls.key \
   --from-file=ca.crt=/tmp/ca.crt
 ```
 
-3) Point to this secret when installing Kuma
+Point to this secret when installing Kuma:
```
lahabana, line 61: You get rid of the 3 but not the 1 and 2 any reason?

docs/docs/dev/security/certificates.md
lahabana: Maybe it should be higher up in the explaination?

docs/docs/dev/security/certificates.md
lobkovilya: I didn't try to get rid of numbers, it's just that 1 is now shared for all environments (kubectl, helm, and universal). Numbers 2 and 3 are environment-specific and that's why they're merged together. I made it look similar to the instructions below on that page.

docs/docs/dev/security/certificates.md
lobkovilya: I added a tip like this:
image

lukidzi (approved):

docs/docs/dev/security/certificates.md
```diff
@@ -28,102 +27,89 @@ Note, this CA is not the same CA for service-to-service communication.
 
 **To override autogenerated certificates**
 
-:::: tabs :options="{ useUrlFragment: false }"
-::: tab "Kubernetes (kumactl)"
+::: tip
+If overridden, Kuma is using the certificates to protect not only [data plane proxy to control plane](#data-plane-proxy-to-control-plane-communication) traffic
```
michaelbeaumont, line 31:
```suggestion
If overridden, Kuma uses the certificates to protect not only [data plane proxy to control plane](#data-plane-proxy-to-control-plane-communication) traffic
```

michaelbeaumont (approved):

docs/docs/dev/security/certificates.md
```diff
@@ -28,102 +27,89 @@ Note, this CA is not the same CA for service-to-service communication.
...
+but also [user to control plane](#user-to-control-plane-communication) traffic
+and [control plane to control plane](#control-plane-to-control-plane-multizone) traffic.
+:::
```
jakubdyszkiewicz, line 34: Let's change `KUMA_DP_SERVER_TLS_CERT_FILE` to `KUMA_GENERAL_TLS_CERT_FILE` then in Universal tab

PR #881, opened 2022-06-27
fix(*): remove default arch to detect correct one

lukidzi:
When the default `ARCH` is set script cannot detect the architecure of the processor.

### DIFF
--- a/docs/.vuepress/public/installer.sh
+++ b/docs/.vuepress/public/installer.sh
@@ -21,7 +21,7 @@
 DIR="$( cd "$( dirname "$0" )" >/dev/null 2>&1 && pwd )"
 
 : "${VERSION:=}"
-: "${ARCH:=amd64}"
+: "${ARCH:=}"
 : "${PRODUCT_NAME:=Kuma}"
 : "${LATEST_VERSION:=https://kuma.io/latest_version}"
 : "${REPO_PREFIX:=kuma}"


slonka (changes requested):
We need add a test to check this does not happen again. What do you think about running `kumactl version` and checking that the [image version](https://github.com/marketplace/actions/run-on-architecture#advanced-example) equals the `kumactl version`?

lukidzi:
> We need add a test to check this does not happen again. What do you think about running `kumactl version` and checking that the [image version](https://github.com/marketplace/actions/run-on-architecture#advanced-example) equals the `kumactl version`?

Added a test which checks if `kumactl version` returns `arm64/amd64`. It requires some tricks because the `kumactl version` call returns an error when connecting to` kuma-cp`, which is not running.

.github/workflows/ci.yml
```diff
@@ -57,3 +57,26 @@ jobs:
...
+
+          # Not required, but speeds up builds by storing container images in
+          # a GitHub package registry.
+          githubToken: ${{ github.token }}
+
+          # Set an output parameter `uname` for use in subsequent steps
+          run: |
+            apt-get update && apt-get install -y curl tar gzip
+            sh docs/.vuepress/public/installer.sh
+            if ./kuma-*/bin/kumactl version | grep arm64 &> /dev/null; then exit 0; else exit 1; fi
```
lahabana, line 82: This is not going to work for future version as the arch shouldn't be in the version to start with.

lukidzi: Changed to `file kumactl` and check which architecture it returns.

.github/workflows/ci.yml
```diff
@@ -52,11 +52,34 @@ jobs:
...
+
+          # Not required, but speeds up builds by storing container images in
+          # a GitHub package registry.
+          githubToken: ${{ github.token }}
+
+          # Set an output parameter `uname` for use in subsequent steps
+          run: |
+            apt-get update && apt-get install -y file curl tar gzip
+            sh docs/.vuepress/public/installer.sh
+            if file ./kuma-*/bin/kumactl | grep aarch64 &> /dev/null; then exit 0; else exit 1; fi
```
michaelbeaumont, line 82:
```suggestion
            file ./kuma-*/bin/kumactl | grep aarch64 &> /dev/null
```

Since it looks like `errexit` and `pipefail` are set here

.github/workflows/ci.yml
```diff
@@ -52,11 +52,34 @@ jobs:
...
-          (apt-get update && apt-get install curl -y) || true
-          yum install tar gzip -y || true
+          (apt-get update && apt-get install file curl -y) || true
+          yum install file tar gzip -y || true
       - uses: actions/checkout@v3
       - name: Run installer.sh
         run: sh docs/.vuepress/public/installer.sh
+      - name: Check kumactl architecture
+        run: file ./kuma-*/bin/kumactl | grep x86-64 &> /dev/null
+  installer-sh-arm64:
```
slonka, line 62: Can't architecture be another dimension in the matrix? We don't have to repeat the `apt-get install` code and it's consistent.

docs/docs/1.7.x/policies/proxy-template.md
```diff
@@ -513,7 +513,7 @@ spec:
 #### HTTP Filter
 
 Modifications that are applied on [HTTP Filters](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/filter/http/http) that are part of [Listeners](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto#config-listener-v3-listener) resource.
-Modifications are applied on all [HTTP Connection Managers](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto.html#http-connection-manager-proto) in the Listener.
+Modifications applies on all [HTTP Connection Managers](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto.html#http-connection-manager-proto) in the Listener.
```
slonka, line 516: This does not sound right I think. Shouldn't this be:

```suggestion
Proxy Template applies modifications on all [HTTP Connection Managers](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto.html#http-connection-manager-proto) in the Listener.
```

It's a bit tricky because the subject is implicit... I don't know, maybe @parkanzky or @michaelbeaumont would know.

docs/docs/dev/policies/proxy-template.md
```diff
@@ -513,7 +513,7 @@ spec:
 #### HTTP Filter
 
 Modifications that are applied on [HTTP Filters](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/filter/http/http) that are part of [Listeners](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto#config-listener-v3-listener) resource.
-Modifications are applied on all [HTTP Connection Managers](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto.html#http-connection-manager-proto) in the Listener.
+Modifications applies on all [HTTP Connection Managers](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto.html#http-connection-manager-proto) in the Listener.
```
slonka, line 516: Same as above.

slonka (changes requested):

docs/docs/1.7.x/policies/proxy-template.md
michaelbeaumont: Wait, why are these changes in this PR at all? :thinking: 

```suggestion
Modifications that Kuma applies on all [HTTP Connection Managers](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto.html#http-connection-manager-proto) in the Listener.
```

Since these are sort of dependent clauses that explain what each modification type is. Here, what `httpFilter` modifications are.

lukidzi: I've changed the URL because it was invalid, and I vale starts to complain, so I had to change it

michaelbeaumont: That was fixed in master I believe, there's no change to the URL here.

.github/workflows/ci.yml
lukidzi: Steps are using different actions. The action used for `arm64` can't be used for `amd64` because of the lack of images supported above. As far as I know, we don't support other architectures so I am not sure if it's worth it.

michaelbeaumont: Agreed, I think it'd make the job more complex to set up the matrix properly and handle selectively running `run-on-arch-action`

slonka (approved):

michaelbeaumont (approved):

PR #883, opened 2022-07-04
feat(install): Argo CD

jakubdyszkiewicz:
xref #2023


slonka:
I'm guessing you wanted to mention this: https://github.com/kumahq/kuma/issues/2023

slonka (commented):
If it's only relevant to `1.7.x` and `dev` then 👍 but I think we should backport it further (e.g. `1.6.x` has `helm` install section - https://kuma.io/docs/1.6.x/installation/helm/ )

PR #903, opened 2022-07-22
docs(kuma-cp) fix tags which are missing "kuma.io/"

parkanzky:
Joe Dascole pointed out that TrafficRoute REST API example was incorrect because it erroneously used `service` in place of `kuma.io/service`. 

Several other examples had the same problem. Dataplane related examples also incorrectly left out the `tags` layer. I tested a random sampling of all these changes before and after, and every one tested required this change.

Signed-off-by: Paul Parkanzky <paul.parkanzky@konghq.com>

### DIFF
--- a/docs/docs/1.6.x/reference/http-api.md
+++ b/docs/docs/1.6.x/reference/http-api.md
@@ -883,7 +883,7 @@ curl -XPUT http://localhost:5681/meshes/mesh-1/dataplanes/backend-1 --data @data
         "port": 11011,
         "servicePort": 11012,
         "tags": {
-          "service": "backend",
+          "kuma.io/service": "backend",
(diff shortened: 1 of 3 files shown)


lahabana:
You haven't fixed dev so this won't get fixed for future versions: The whole backporting tool to have changes in multiple versions is described there: https://github.com/kumahq/kuma-website/blob/master/WRITING-DOCUMENTATION.md#fixing-typos

michaelbeaumont (commented):
~Shouldn't this be in `docs/dev` as well?~ Hadn't seen Charly's comment

lahabana:
@parkanzky can we try to push that through?

michaelbeaumont (approved):
Good improvement

PR #908, opened 2022-07-25
fix(ipv6): add doc to disable ipv6 support

lahabana:


michaelbeaumont (approved):

michaelbeaumont:
@lahabana wasn't there a discussion recently about whether we need a separate redirect port for ipv6? Would that deprecate this annotation?

lahabana:
@michaelbeaumont yes there were and this will be deprecated yes. We'll also need to introduce an alternative `disableIpv6` flag or something like this in the transparentProxy section of the dataplane. The issue for deprecating is: kumahq/kuma#4562

PR #914, opened 2022-07-27
fix(sync_generated.sh): add proper remote handling

michaelbeaumont:
Otherwise we can get errors with multiple branches:

```
fatal: 'release-1.1' matched multiple (2) remote tracking branches
```

### DIFF
--- a/sync_generated.sh
+++ b/sync_generated.sh
@@ -1,5 +1,19 @@
 #!/bin/bash
 set -e
+
+pushd ../kuma
+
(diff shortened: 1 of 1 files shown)


lahabana (approved):
Was that happening locally? This script usually runs as a github action: https://github.com/kumahq/kuma-website/blob/master/.github/workflows/sync-generated.yml

michaelbeaumont:
Ah true. The error in the description was happening due to there being more than one remote.

PR #938, opened 2022-08-02
docs(kuma-cp): changes to retry policy

johnharris85:
Signed-off-by: John Harris <john@johnharris.io>

Adds docs for https://github.com/kumahq/kuma/pull/4744


docs/docs/dev/policies/retry.md
```diff
@@ -136,11 +136,42 @@ We will apply the configuration with `kumactl apply -f [..]` or via the [HTTP AP
   ```
 
   :::tip
-  Note that if you won't provide `retriableStatusCodes`, the default behaviour of the policy is to retry:
-  - when server responds with one of status codes: `502`, `503` or `504`,
-  - when server won't respond at all (disconnect/reset/read timeout),
-  - when server resets the stream with a `REFUSED_STREAM` error code.
+  If both `retriableStatusCodes` is provided in addition to `retryOn` (below), but the latter doesn't contain `retriable_status_codes` as a condition, it will be automatically added.
   :::
+
```
michaelbeaumont, line 141: IMO we should try to link to the reference docs instead of explaining all the options in the policy pages and trying to keep them up to date. Can we do that here?

johnharris85: We can. Was just following the existing GRPC retry list convention, but I agree in principle. One thing to consider though (and there may be a way to solve this in the code) is that we have to modify the option `5xx` as protobuf enum fields can't start with a non-alpha character. Hence it's `all_5xx` here and we'd need to document the delta between us and the envoy options.

michaelbeaumont: OK, good point. We can probably manage to display comments for individual enums, there's definitely still some work to do on the generated docs.

docs/docs/dev/policies/retry.md
```diff
@@ -136,11 +136,42 @@ We will apply the configuration with `kumactl apply -f [..]` or via the [HTTP AP
...
+  - `http3_post_connect_failure`
+
+  :::tip
+  Note that if `retryOn` is not defined or if it's empty, the policy will default to the equivalent of:
+
+  ```yaml
+  retryOn:
+  - gateway-error
+  - connect-failure
+	- refused-stream
```
michaelbeaumont, line 166: could cause trouble for copy/paste

```suggestion
  - refused-stream
```

michaelbeaumont (approved):

PR #943, opened 2022-08-04
fix(referencing): fix sitemap to exclude other pages

lahabana:
- fix sitemap to only include the version of the latest release
- do a few page performance improvements


michaelbeaumont (approved):
:+1:

PR #969, opened 2022-08-19
docs(gateway): fix absolute path

slonka:

fix absolute path that would lead to new versions linking to old versions of the docs


kleinfreund (approved):
Yeah, that looks right. Should this also be fixed in `docs/docs/1.8.x/quickstart/kubernetes.md`? There are three other (but old) instances of absolute URLs in the docs (searching the project for `](https://kuma.io/docs/1.`), but that could be intentional.

slonka:
> Yeah, that looks right. Should this also be fixed in docs/docs/1.8.x/quickstart/kubernetes.md?

Yes, but this in `master` right now.

PR #976, opened 2022-08-23
fix: access control headers missing allowed method and headers

kleinfreund:
Adds the `Access-Control-Allow-Method` and `Access-Control-Allow-Headers` headers to the `/latest_version` URL path so that preflight requests include them and CORS requests with a set `Content-Type` are allowed.


kleinfreund:
It’s working. Test by running `curl -i -X OPTIONS https://kuma.io/latest_version | grep 'access-control-allow-'`

```
access-control-allow-headers: Content-Type
access-control-allow-method: GET
access-control-allow-origin: *
```

PR #986, opened 2022-08-31
fix(build): disable limit chunk count plugin in dev

slonka:
Disable limit chunk count plugin in dev to speed up rebuild times.

First build still takes a lot of time but subsequent rebuilds are much faster now:

previously:
```
success [10:38:00] Build 878092 finished in 28634 ms! ( http://localhost:8080/ )
```

now:

```
success [10:53:50] Build 07ff8b finished in 1461 ms! ( http://localhost:8080/ )
```

I've figured it out by recording a flamegraph and seeing that we spend a lot of time there:

image

If you want original flamegraph or instructions on how to do this let me know.

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -511,12 +511,18 @@ Sitemap: https://kuma.io/sitemap.xml
   evergreen: false,
   configureWebpack: (config) => {
     return {
-      plugins: [
-        new webpack.EnvironmentPlugin({...process.env}),
-        new webpack.optimize.LimitChunkCountPlugin({
-          maxChunks: 20
-        })
-      ]
+      plugins: (function() {
+        const sharedPlugins = [
+          new webpack.EnvironmentPlugin({...process.env}),
+        ];
+        if (process.env.NODE_ENV.toLowerCase() === "production") {
+          return [...sharedPlugins, new webpack.optimize.LimitChunkCountPlugin({
+            maxChunks: 20
+          })]
+        } else {
+          return sharedPlugins;
+        }
+      })()
     }
   },
   shouldPrefetch: false,


bartsmykla (approved):
lgtm

lahabana (commented):
HAve we considered removing the limitChunkCountPlugin altogether?

slonka:
> Have we considered removing the limitChunkCountPlugin altogether?

I think this should be enabled in prod, right @kleinfreund? we don't want to build too many files that will then add HTTP overhead to load times.

kleinfreund:
The plugin (https://webpack.js.org/plugins/limit-chunk-count-plugin/) limits the amount of separate JavaScript chunks webpack is allowed to generate. Specifically, this influences the amount of files generated in `docs/.vuepress/dist/assets/js`.

Currently, 19 JavaScript files are being generated. Disabling this plugin changes this figure drastically. Without, we generate 1354 JavaScript files. This could have noticeable negative impact on the site’s performance because it will increase the amount of network requests the site has to trigger to fetch the necessary JS chunks.

For the time being, I would rather experiment with the proposed change of only disabling it during development.

lahabana:
> Currently, 19 JavaScript files are being generated. Disabling this plugin changes this figure drastically. Without, we generate 1354 JavaScript files. This could have noticeable negative impact on the site’s performance because it will increase the amount of network requests the site has to trigger to fetch the necessary JS chunks.

Right but here's the lighthouse output:
image


Looks like this limit of 20 chunks is affecting more negatively perf than positively

kleinfreund:
@lahabana That’s a fair point. It is true that loading bigger chunks, while reducing the overall network requests (depends on scenario) also has a memory penalty. I’m definitely happy try and see if we can get rid of the chunk limit.

lahabana:
So what's next here?

kleinfreund:
I’m happy to try it with removing the plugin completely. What do you think, @slonka?

kleinfreund:
@slonka Could you perhaps add a commit to this PR’s branch that removes the plugin so we can run checks against the Netlify preview demo? I want to see how many of the chunks would be loaded on a few pages. For example, currently, going to kuma.io, you can see that we load 17 chunks. If we find that loading the same few pages without chunk limiting also results in a majority of the chunks being loaded, I don’t think the plugin can be disabled. But if this does actually end up only loading some of them (and consequently cause less data to be transferred over the wire), it might be worth the trade off (that being one of amount of network requests versus amount of transferred data).

slonka:
Yeah sure, I'll do that :) also feel free to edit this PR if you want to, it has "allow maintainers to edit" :)

slonka:
image

vs 

image

kleinfreund:
Yeah, it does load all the chunks right away. They’re bigger too in total, presumably because of some boiler plate export code.

(You can filter these requests by selecting the "Other" category)

So, let’s keep it to your initial proposal?

slonka:
> So, let’s keep it to your initial proposal?

Yeah, I'd go with that. Just one thing: are we sure, we're setting `NODE_ENV=production` on netlify?

kleinfreund:
@slonka Considering that we use `process.env.NODE_ENV === 'production'` in a couple of places, I reckon that happens. That being said, you can drop the `.toLowerCase()`. I’ve never seen that capitalized.

PR #987, opened 2022-08-31
docs(gateway): clarify behavior of `MeshGatewayInstance`

### LINKED ISSUE #884: Explore/Gateway/Builtin page gives wrong impression about the behaviour
### What happened?

From docs:

> ```
> echo "
> apiVersion: kuma.io/v1alpha1
> kind: MeshGatewayInstance
> metadata:
>   name: edge-gateway
> spec:
>   replicas: 1
>   serviceType: LoadBalancer
>   tags:
>     kuma.io/service: edge-gateway
> " | kubectl apply -f -
> ```
> In the example above, the control plane will create a new Deployment in the gateways namespace. This deployment will have the requested number of builtin gateway Dataplane pod replicas, which will be configured as part of the service named in the MeshGatewayInstance tags.

While in reality a new Deployment won't be created until there is a MeshGateway object.

michaelbeaumont:
Closes #884 and fixes some vale suggestions/errors


bartsmykla (approved):
lgtm

PR #992, opened 2022-09-03
Install was always default to latest

MustkimKhatik:
Signed-off-by: MustkimKhatik <83959074+MustkimKhatik@users.noreply.github.com>

Install was always pointing to the latest version because of  ```curl -L https://kuma.io/installer.sh | bash - ``` that what was the ```install_os``` having, so its changed to ```curl -L https://kuma.io/installer.sh | VERSION={{ $page.latestVersion }} sh -``` 

This change is done inside the ```div``` class because the normal fenced code was preventing the interpolation of ```{{$page.latestVersion }}``` in the documents.

### DIFF
--- a/docs/.vuepress/_snippets/install_kumactl.md
+++ b/docs/.vuepress/_snippets/install_kumactl.md
@@ -5,7 +5,9 @@ To run Kuma on Kubernetes, you need to download the Kuma cli (`kumactl`) on your
 
 You can run the following script to automatically detect the operating system and download Kuma:
 
-`curl -L https://kuma.io/installer.sh | VERSION={{ $page.latestVersion }} bash -`
+<div class="language-sh">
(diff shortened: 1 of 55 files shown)


MustkimKhatik:
Hey @kleinfreund, you can check here instead.

MustkimKhatik:
For preview- https://6312e6766702a751e889d93a--incandescent-shortbread-3cf54f.netlify.app

kleinfreund (approved):
Thank you for your contribution. :) This is looking good to me now.

MustkimKhatik:
> Thank you for your contribution. :) This is looking good to me now.

Alright! 🙌

slonka:
If @kleinfreund is ok with it then let's 🚢 this 👍

slonka (approved):

PR #1021, opened 2022-10-05
fix: make it explicit that gateway-api only works in standalone

lahabana:


docs/docs/dev/explore/gateway-api.md
```diff
@@ -6,6 +6,8 @@ Kuma supports configuring [built-in gateway](gateway.md) using [Kubernetes Gatew
 
 ::: warning
 [Kubernetes Gateway API](https://gateway-api.sigs.k8s.io/) is still beta, therefore Kuma's integration provides the same level of stability.
+
+gateway-api is not supported in multi-zone. To do this you need to use [Kuma MeshGateway](../gateway.md)
```
michaelbeaumont, line 10:
```suggestion
Gateway API is not supported in multi-zone. To use the builtin Gateway, you need to use the [`MeshGateway`  resources](../gateway.md).
```

Issue #1015, opened 2022-09-30
Redirect all kuma.io annotations to their documentation


brentos:
### Description

kuma.io annotations should forward to their documentation, for example:

When I search for [kuma.io/sidecar-env-vars](http://kuma.io/sidecar-env-vars) , chrome sends me to that URL and  i get a 404 because it doesn't exist; instead it would be nice if it redirected me to https://kuma.io/docs/1.8.x/reference/kubernetes-annotations/#kuma-io-sidecar-env-vars

jakubdyszkiewicz:
> Triage: we would also need to handle subdomains (`k8s.kuma.io/...` and `traffic.kuma.io/...`)

slonka:
We are still missing redirects for subdomains AFAICT.

slonka:
Subdomains still don't work e.g. http://prometheus.metrics.kuma.io/port

bartsmykla:
Still the case as it seems, these doesn't work:
https://github.com/kumahq/kuma-website/blob/f651f77bb0426a56b060efc84db0623214f4a679/app/_redirects#L37-L41

PR #1032, opened 2022-10-11
Redirect kuma.io annotations to their appropriate documents

MustkimKhatik:
This PR adds [#1015](https://github.com/kumahq/kuma-website/issues/1015#issue-1393002325) feature


MustkimKhatik:
Hey @slonka, could you check this?

docs/.vuepress/enhanceApp.js
```diff
@@ -64,6 +64,106 @@ export default ({
...
+    {
+      name: 'transparent-proxying-inbound-v6-port',
+      path: '/transparent-proxying-inbound-v6-port',
+      redirect: "https://kuma.io/docs/1.8.x/reference/kubernetes-annotations/#kuma-io-transparent-proxying-inbound-v6-port"
+    },
+    {
+      name: 'sidecar-drain-time',
+      path: '/sidecar-drain-time',
+      redirect: "https://kuma.io/docs/1.8.x/reference/kubernetes-annotations/#kuma-io-sidecar-drain-time"
+    }
```
kleinfreund, line 167: **Suggestion**: I think the redirection should happen via the Netlify redirect rules that are defined here:

https://github.com/kumahq/kuma-website/blob/3fe10d33405c93d69a2c5b50c0499008a114d16f/docs/.vuepress/config.js#L343-L407

This way the redirect happens before loading the Vuepress application on the client side.

I think the redirects should also dynamically determine what’s the latest stable release so we don’t have to manually update the version string every time we release a new stable version.

kleinfreund (changes requested):

docs/.vuepress/enhanceApp.js
slonka: > I think the redirects should also dynamically determine what’s the latest stable release 

For this **I think** we can use `latest` e.g. https://kuma.io/docs/latest/policies/traffic-log/#add-a-logging-backend redirects to https://kuma.io/docs/1.8.x/policies/traffic-log/#add-a-logging-backend

MustkimKhatik: Alright I'll apply changes accordingly

slonka:
one question remains: how do we keep this in sync with [this](https://github.com/kumahq/kuma/blob/master/pkg/plugins/runtime/k8s/metadata/annotations.go#L13)? maybe we should put a link in the description and do some sort of grepping / code generation?

docs/.vuepress/enhanceApp.js
MustkimKhatik: This works now but there are some paths that are not following the common redirection syntax like [this one](https://github.com/kumahq/kuma-website/pull/1032/commits/62937958a0408a3e122b1317111005183137a753#:~:text=%60prometheus.metrics.kuma.io/port%20/docs/%24%7Bversions.latestMinor%7D/reference/kubernetes%2Dannotations/%23prometheus%2Dmetrics%2Dkuma%2Dio%2Dport%60%2C) and I can't actually test this because it showing me some privacy-related issue in the browser while testing locally. Is this still right or have some changes in it?

MustkimKhatik:
Hey @slonka, can you please elaborate on your suggestion?

docs/.vuepress/config.js
```diff
@@ -370,6 +370,34 @@ Sitemap: https://kuma.io/sitemap.xml
...
+            `/service-account-token-volume /docs/${versions.latestMinor}/reference/kubernetes-annotations/#kuma-io-service-account-token-volume`,
+            `/transparent-proxying-reachable-services /docs/${versions.latestMinor}/reference/kubernetes-annotations/#kuma-io-transparent-proxying-reachable-services`,
+            `prometheus.metrics.kuma.io/aggregate-<name>-enabled /docs/${versions.latestMinor}/reference/kubernetes-annotations/#prometheus-metrics-kuma-io-aggregate-name-enabled`,
+            `prometheus.metrics.kuma.io/aggregate-<name>-path /docs/${versions.latestMinor}/reference/kubernetes-annotations/#prometheus-metrics-kuma-io-aggregate-name-path`,
+            `prometheus.metrics.kuma.io/aggregate-<name>-port /docs/${versions.latestMinor}/reference/kubernetes-annotations/#prometheus-metrics-kuma-io-aggregate-name-port`,
+            `/transparent-proxying-inbound-v6-port /docs/${versions.latestMinor}/reference/kubernetes-annotations/#kuma-io-transparent-proxying-inbound-v6-port`,
+            `/sidecar-drain-time /docs/${versions.latestMinor}/reference/kubernetes-annotations/#kuma-io-sidecar-drain-time`,
```
kleinfreund, line 400: **Suggestion**: This is the right place now, but I’d like to suggest the following improvement: instead of defining the redirects inline right here, reference an array like so:

```suggestion
            ...kumaIoTagRedirects,
```

and then above (beneath where the `policyRedirects` array is defined), you add the following definition. This saves some repetition of common path segments and is easier to follow.

```javascript
          // Redirects known “kuma.io/*” tags to their respective page in the docs.
          const kumaIoTagRedirects = [
            '/sidecar-injection', '#labels',
            '/mesh', '#kuma-io-mesh',
            '/sidecar-injection', '#kuma-io-sidecar-injection-2',
            '/gateway', '#kuma-io-gateway',
            '/ingress', '#kuma-io-ingress',
            '/ingress-public-address', '#kuma-io-ingress-public-address',
            '/ingress-public-port', '#kuma-io-ingress-public-port',
            '/direct-access-services', '#kuma-io-direct-access-services',
            '/virtual-probes', '#kuma-io-virtual-probes',
            '/virtual-probes-port', '#kuma-io-virtual-probes-port',
            '/sidecar-env-vars', '#kuma-io-sidecar-env-vars',
            '/container-patches', '#kuma-io-container-patches',
            'prometheus.metrics.kuma.io/port', '#prometheus-metrics-kuma-io-port',
            'prometheus.metrics.kuma.io/path', '#prometheus-metrics-kuma-io-path',
            '/builtindns', '#kuma-io-builtindns',
            '/builtindnsport', '#kuma-io-builtindnsport',
            '/ignore', '#kuma-io-ignore',
            'traffic.kuma.io/exclude-inbound-ports', '#traffic-kuma-io-exclude-inbound-ports',
            'traffic.kuma.io/exclude-outbound-ports', '#traffic-kuma-io-exclude-outbound-ports',
            '/transparent-proxying-experimental-engine', '#kuma-io-transparent-proxying-experimental-engine',
            '/envoy-admin-port', '#kuma-io-envoy-admin-port',
            '/service-account-token-volume', '#kuma-io-service-account-token-volume',
            '/transparent-proxying-reachable-services', '#kuma-io-transparent-proxying-reachable-services',
            'prometheus.metrics.kuma.io/aggregate-<name>-enabled', '#prometheus-metrics-kuma-io-aggregate-name-enabled',
            'prometheus.metrics.kuma.io/aggregate-<name>-path', '#prometheus-metrics-kuma-io-aggregate-name-path',
            'prometheus.metrics.kuma.io/aggregate-<name>-port', '#prometheus-metrics-kuma-io-aggregate-name-port',
            '/transparent-proxying-inbound-v6-port', '#kuma-io-transparent-proxying-inbound-v6-port',
            '/sidecar-drain-time', '#kuma-io-sidecar-drain-time',
          ].map(([sourcePath, tagValue]) => `${sourcePath}/ /docs/${versions.latestMinor}/reference/kubernetes-annotations/${tagValue}/ 301`)
```

kleinfreund (changes requested):
Hey. Thanks again for your contribution. I just have a suggestion for an improvement.

As for the suggestion made by @slonka in https://github.com/kumahq/kuma-website/pull/1032#issuecomment-1276340799, I think this is something we could consider follow-up work. It would be nice to have these redirects be based on the actual list of known kuma.io/* tags because then we wouldn’t have to remember to manually keep this list up to date. We’ll discuss this.

docs/.vuepress/enhanceApp.js
kleinfreund: @MustkimKhatik I’ve addressed your question in https://github.com/kumahq/kuma-website/pull/1032#discussion_r995438003.

docs/.vuepress/config.js
MustkimKhatik: Alright, @kleinfreund I applied the changes, and thank you for the guidance!

kleinfreund: @MustkimKhatik Thanks! Now you’re just missing `kumaIoTagRedirects` in the list of redirects that are actually applied (e.g. `...kumaIoTagRedirects,` beneath `...policyRedirects,`).

MustkimKhatik: Yes, it's done @kleinfreund!

kleinfreund: Thank you. That’s it. I’ll talk to folks and see whether we can merge this already or whether we want to wait and come up with a sensible strategy for keeping this list of redirects up-to-date with the annotations in Kuma (see https://github.com/kumahq/kuma/blob/master/pkg/plugins/runtime/k8s/metadata/annotations.go#L13).

MustkimKhatik: Okay sure @kleinfreund

slonka: > you won’t just be able to type prometheus.metrics.kuma.io/path into the address bar and find the right docs via redirects because prometheus.metrics.kuma.io isn’t actually a served origin

It's not possible to redirect all subdomains in netlify but we can redirect them one by one: https://answers.netlify.com/t/redirect-all-subdomains-to-main-domain/4804 & https://docs.netlify.com/routing/redirects/ - I'd keep those entries and redirect all subdomains that we know.

kleinfreund: If we really want to extend this to subdomains, I suggest a second array `kumaIoSubdomainRedirects` because the redirect would be constructed in a different manner:

```javascript
const kumaIoSubdomainRedirects = [
  ['https://prometheus.metrics.kuma.io/port', `/docs/${versions.latestMinor}/reference/kubernetes-annotations/#prometheus-metrics-kuma-io-port`],
  ['https://prometheus.metrics.kuma.io/path', `/docs/${versions.latestMinor}/reference/kubernetes-annotations/#prometheus-metrics-kuma-io-path`],
  ['https://traffic.kuma.io/exclude-inbound-ports', `/docs/${versions.latestMinor}/reference/kubernetes-annotations/#traffic-kuma-io-exclude-inbound-ports`],
  ['https://traffic.kuma.io/exclude-outbound-ports', `/docs/${versions.latestMinor}/reference/kubernetes-annotations/#traffic-kuma-io-exclude-outbound-ports`],
].map(([sourceUrl, destinationPath]) => `${sourceUrl} ${destinationPath}/ 301`)
```

I’m leaving those entries out that that have dynamic dynamic segments (e.g. `^prometheus.metrics.kuma.io/aggregate-([a-zA-Z0-9-]+)-(port|path|enabled)$`).

slonka: agree 👍

MustkimKhatik: I'd simply add this one then

kleinfreund: Very nice, thank you. I’ll consider this done now. :)

kleinfreund (approved):
That’s a nice addition to make very specific docs more discoverable. Thank you for your contribution (again)!

kleinfreund:
**Comment** (cc @slonka): Is the linker checker workflow flaky? I’m seeing this fail a lot and I can never explain why.

slonka:
It's the least flaky that we tried 😑 but there is some still

MustkimKhatik:
> That’s a nice addition to make very specific docs more discoverable. Thank you for your contribution (again)!

Sure @kleinfreund, happy to be a part of it

slonka:
> **Comment** (cc @slonka): Is the linker checker workflow flaky? I’m seeing this fail a lot and I can never explain why.

bumping muffet to [v2.6.1](https://github.com/raviqqe/muffet/releases/tag/v2.6.1) does not help (tested locally) - also created an issue [there](https://github.com/raviqqe/muffet/issues/254). I think we could:
1. disable link checker as a "required" check and just make sure that there is not a ton of issues / periodically manually check if nothing changed
2. add this site to exceptions

WDYT?

slonka:
Why did this get merged? Subdomains do not work, I pointed out [here](https://github.com/kumahq/kuma-website/pull/1032#discussion_r996732970) - that we can follow this - https://docs.netlify.com/routing/redirects/ to add redirects.

kleinfreund:
@slonka Because all the changes that can be done in code to support domain-level redirections were added (see also the very thread you linked to: https://github.com/kumahq/kuma-website/pull/1032#discussion_r996746240). The redirections don’t work, as far as I understand anyway, because they need to be explicitly configured in Netlify (see https://docs.netlify.com/routing/redirects/redirect-options/#domain-level-redirects).

slonka:
> Because all the changes that can be done in code to support domain-level redirections were added

I don't think this is the case. If I'm not mistaken you put redirects in a file called "netlify.toml" or "_redirects" in your repo. [Example](https://axway-open-docs.netlify.app/): https://github.com/Axway/axway-open-docs/blob/cbcad19497edbc363387927c67b30cd82b58eb6a/netlify.toml#L27

slonka:
So why isn't this working? Is DNS not set up properly?

kleinfreund:
@slonka Yes, this needs explicit DNS configuration.

Issue #1017, opened 2022-10-04
Write MeshTrace docs


slonka:
### Description

After https://github.com/kumahq/kuma/pull/5085 is merged we need to create docs for this new Policy.

PR #1034, opened 2022-10-17
feat(kumacp): mesh trace docs

slonka:
Closes https://github.com/kumahq/kuma-website/issues/1017

### DIFF
--- a/docs/docs/dev/policies/matching.md
+++ b/docs/docs/dev/policies/matching.md
@@ -1,5 +1,5 @@
-# Matching matrix
-
+# Policy Matching
+## Matrix
 | TargetRef type    | top level | to  | from |
 |-------------------|-----------|-----|------|
 | Mesh              | ✅         | ✅   | ❌    |
@@ -48,10 +48,12 @@ A combination of a policy and the top level `targetRef` can influence which type
 If a policy does not support a level then every type will be marked with ❌ in that level
 and the YAML configuration must omit that field from the definition.
 
-# Matching meaning
+# targetRef levels
 
 Top level `targetRef` defines which set of proxies a policy will affect.
-`To` level `targetRef` defines rules for outgoing traffic relative to the top level.
+`Top` level `targetRef` defines which set of proxies a policy will affect (.i.e: the set of proxies whose configuration is getting modified).
+`To` level `targetRef` defines rules that applies to outgoing traffic of proxies selected by the `Top` level `targetRef`.
+`From` level `targetRef` defines rules that applies to incoming traffic of proxies selected by the `Top` level `targetRef`.
 `From` level `targetRef` defines rules for incoming traffic relative to the top level.
 
 Consider the example below:
@@ -90,6 +92,6 @@ spec:
               path: '/tmp/logs.txt'
 ```
 
-This policy will target a service `web-frontend`.
-It instructs Kuma to log traffic **coming** from `web-backend` to `web-frontend`.
-It also instructs to log traffic going out of `web-frontend` to **anything** in the `Mesh` named `default`.
+This policy will target all proxies that implement the service `web-frontend`.
+It logs traffic **coming** from `web-backend` to `web-frontend`.
+It also logs traffic going out of proxies that implement `web-frontend` to **anything** in the `Mesh`.
--- a/docs/docs/dev/policies/matching.md
+++ b/docs/docs/dev/policies/matching.md
@@ -48,7 +48,7 @@ A combination of a policy and the top level `targetRef` can influence which type
 If a policy does not support a level then every type will be marked with ❌ in that level
 and the YAML configuration must omit that field from the definition.
 
-# targetRef levels
+## targetRef levels
 
 Top level `targetRef` defines which set of proxies a policy will affect.
 `Top` level `targetRef` defines which set of proxies a policy will affect (.i.e: the set of proxies whose configuration is getting modified).
--- a/docs/docs/dev/policies/matching.md
+++ b/docs/docs/dev/policies/matching.md
@@ -1,0 +1,95 @@
+# Matching matrix
+
+| TargetRef type    | top level | to  | from |
+|-------------------|-----------|-----|------|
+| Mesh              | ✅         | ✅   | ❌    |
+| MeshSubset        | ✅         | ❌   | ❌    |
+| MeshService       | ✅         | ❌   | ✅    |
+| MeshServiceSubset | ✅         | ❌   | ❌    |
+| MeshGatewayRoute  | ❌         | ❌   | ❌    |
+| MeshHTTPRoute     | ❌         | ❌   | ❌    |
+
+The above matrix indicates which entity types (`Mesh`, `MeshSubset`, `MeshService`, `MeshServiceSubset`, `MeshGatewayRoute`, `MeshHTTPRoute`) a policy supports
+and where they can be placed (top level, to level, from level).
+
+Looking at the structure of the `spec` field of a policy, `targetRef` can be used in multiple places:
+
+```yaml
+apiVersion: kuma.io/v1alpha1
+kind: PolicyType
+...
+spec:
+  targetRef: # top level
+    kind: Mesh | MeshSubset | MeshService | MeshServiceSubset
+    # ...
+  to:
+    - targetRef: # "to"
+        kind: Mesh
+      # ...
+  from:
+    - targetRef: # "from"
+        kind: MeshService
+      # ...
+```
+
+and the table defines which type is available in each place.
+
+In this example:
+- top level can be one of `Mesh`, `MeshSubset`, `MeshService`, `MeshServiceSubset`
+- to level can only be `Mesh`
+- from level can only be `MeshService`
+
+A policy might not support every level
+(e.g. `MeshTrace` only allows top level `targetRef`).
+
+A combination of a policy and the top level `targetRef` can influence which types are available in `to` and `from` levels
+(e.g. `MeshAccessLog` with a top level `targetRef` of `MeshGatewayRoute` can only have `from` level because there is no outbound listener).
+
+If a policy does not support a level then every type will be marked with ❌ in that level
+and the YAML configuration must omit that field from the definition.
+
+# Matching meaning
+
+Top level `targetRef` defines which set of proxies a policy will affect.
+`To` level `targetRef` defines rules for outgoing traffic relative to the top level.
+`From` level `targetRef` defines rules for incoming traffic relative to the top level.
+
+Consider the example below:
+
+```yaml
+apiVersion: kuma.io/v1alpha1
+kind: MeshAccessLog
+metadata:
+  name: example
+  namespace: kuma-system
+  labels:
+    kuma.io/mesh: default
+spec:
+  targetRef: # top level targetRef
+    kind: MeshService
+    name: web-frontend
+  to:
+    - targetRef: # to level targetRef
+        kind: MeshService
+        name: web-backend
+      default:
+        backends:
+          - file:
+              format:
+                plain: '{"start_time": "%START_TIME%"}'
+              path: '/tmp/logs.txt'
+  from:
+    - targetRef: # from level targetRef
+        kind: Mesh
+        name: default
+      default:
+        backends:
+          - file:
+              format:
+                plain: '{"start_time": "%START_TIME%"}'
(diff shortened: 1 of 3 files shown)


docs/docs/dev/policies/mesh-trace.md
```diff
@@ -0,0 +1,340 @@
+# Mesh Trace (beta)
```
slonka, line 1: Would love feedback on how to mark that this policy is not ready to be used & clashes with TrafficTrace

docs/docs/dev/policies/mesh-trace.md
```diff
@@ -0,0 +1,340 @@
...
+
+```yaml
+sampling:
+  overall:
+    value: 80
+  random:
+    value: 60
+  client:
+    value: 40
+```
```
lahabana, line 310: Please say which one is the most likely useful :)

docs/docs/dev/policies/mesh-trace.md
```diff
@@ -0,0 +1,340 @@
...
+::: tab "Full example"
+
+```yaml
+apiVersion: kuma.io/v1alpha1
+kind: MeshTrace
+metadata:
+  name: default
+  namespace: kuma-system
+  labels:
+    kuma.io/mesh: default
```
lahabana, line 173:
```suggestion
    kuma.io/mesh: default # optional, defaults to `default` if unset
```

docs/docs/dev/policies/mesh-trace.md
lahabana: I'm fine with this as it matches Kong's policy: https://docs.konghq.com/gateway/latest/stability/#main

docs/docs/dev/policies/mesh-trace.md
```diff
@@ -0,0 +1,340 @@
...
+    sampling:
+      overall:
+        value: 80
+      random:
+        value: 60
+      client:
+        value: 40
+```
+:::
+::: tab "Minimal example"
```
lahabana, line 76: Probably have minimal first

docs/docs/dev/policies/mesh-trace.md
```diff
@@ -0,0 +1,340 @@
...
+For `header` - if a value is missing then `default` will be used.
+If `default` is not provided then the tag won't be added.
+
+## Targeting parts of the infrastructure
+
+With a lot of traffic going through the system,
+a valid use case is to target specific set of microservices that are part of a main request path
+(e.g. a process of buying an item in an e-commerce application).
+
+You can achieve that by using more specific `targetRef` and (if needed) a separate backend.
```
lahabana, line 340:
```suggestion
You can achieve that by using a more specific `targetRef` and (if needed) a separate backend.
```

docs/docs/dev/policies/mesh-trace.md
```diff
@@ -0,0 +1,340 @@
...
+```
+
+For `header` - if a value is missing then `default` will be used.
+If `default` is not provided then the tag won't be added.
+
+## Targeting parts of the infrastructure
+
+With a lot of traffic going through the system,
+a valid use case is to target specific set of microservices that are part of a main request path
+(e.g. a process of buying an item in an e-commerce application).
```
lahabana, line 338: I think this paragraph is not super clear.

I think this to make sense you really need to add an example.

docs/docs/dev/policies/mesh-trace.md
slonka: WDYT about this nested style? I wanted to have a way to show the minimal example (something that people can start with and there is not too much tinkering) and the full example so it's easy to figure out what's possible. Do you think it's worth having this structure as a "template" for future policies?

docs/docs/dev/policies/mesh-trace.md
slonka: changed to be the mirror of the "tip" in old docs

lahabana:
@slonka I was thinking about something last night.
It would be nice to have a standard table or something like this at the top of each policy that sums up some of the feature for example:

|use To  targetRef | use From targetRef | Works with builtinGateway | Works with externalService |
|-------|----------| ---------------------------| ----------------------------|
| x.        | x             | x                                             | x |


This is just an idea but I feel it would help users quickly understand a policy

slonka:
@lahabana - yeah, that's a great idea, but I'd love some clarification on a couple of  things:
- "use To targetRef" - do you mean that you can use "To" in this policy, or do you think we should list which "targets" (Mesh, MeshSubset, Service)... are available?
- is "Works with builtinGateway" equivalent to being able to target `MeshGatewayRoute`? Why is `builtinGateway` singled out here, we could do tracing on delegated gateway right? 
- "Works with externalService" - we don't have a specific target for external / internal services, it's all `MeshService` right?

I think a table like this would be very nice but I'd structure it a bit differently:

MeshTrace:

| `targetRef`            | `to.targetRef` | `from.targetRef` |
|------------------------|----------------|------------------|
| Mesh                   | x              | x                |
| MeshSubset             |                |                  |
| MeshService            |                |                  |
| MeshServiceSubset      |                |                  |
| MeshGatewayRoute       |                |                  |

---

MeshAccessLog:

| `targetRef`            | `to.targetRef` | `from.targetRef` |
|------------------------|----------------|------------------|
| Mesh                   | Mesh           | Mesh             |
| MeshSubset             | MeshService    |                  |
| MeshService            |                |                  |
| MeshServiceSubset      |                |                  |
| MeshGatewayRoute       |                |                  |
| MeshHTTPRoute          |                |                  |

WDYT?

lahabana:
Indeed we need to indicate this:

| TargetRef type | top level | to | from   |
|---------------|----------| ---| ------ |
| Mesh | x |  | |
| MeshSubset |  | x | |


Feels like this might be a bit more readable.

slonka:
> Feels like this might be a bit more readable.

agree! 👍 - for now I'll just create a table (since there is a migration to jekyll and creating a reusable component in vuepress might mean we have to re-do the work)

lahabana:
Agreed let's stick with markdown for now

docs/docs/dev/policies/mesh-trace.md
```diff
@@ -290,7 +303,7 @@ By default, this property is set to false.
 ### Sampling
 
 You can configure sampling settings equivalent to Envoy's:
-- [overall](https://www.envoyproxy.io/docs/envoy/v1.22.5/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto.html?highlight=overall_sampling#extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-tracing)
+- [overall](https://www.envoyproxy.io/docs/envoy/v1.22.5/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto.html?highlight=overall_sampling#extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-tracing) - the most commonly used
 - [random](https://www.envoyproxy.io/docs/envoy/v1.22.5/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto.html?highlight=random_sampling#extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-tracing)
 - [client](https://www.envoyproxy.io/docs/envoy/v1.22.5/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto.html?highlight=client_sampling#extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-tracing)
```
lahabana, line 308: I'd go with:

```
tip::
Most of the time setting only `overall` is sufficient; `random` and `client` are for advanced use cases.
```

docs/docs/dev/policies/matching.md
```diff
@@ -0,0 +1,95 @@
...
+A combination of a policy and the top level `targetRef` can influence which types are available in `to` and `from` levels
+(e.g. `MeshAccessLog` with a top level `targetRef` of `MeshGatewayRoute` can only have `from` level because there is no outbound listener).
+
+If a policy does not support a level then every type will be marked with ❌ in that level
+and the YAML configuration must omit that field from the definition.
+
+# Matching meaning
+
+Top level `targetRef` defines which set of proxies a policy will affect.
+`To` level `targetRef` defines rules for outgoing traffic relative to the top level.
```
lahabana, line 54:
```suggestion
`Top` level `targetRef` defines which set of proxies a policy will affect (.i.e: the set of proxies whose configuration is getting modified).
`To` level `targetRef` defines rules that applies to outgoing traffic of proxies selected by the `Top` level `targetRef`.
`From` level `targetRef` defines rules that applies to incoming traffic of proxies selected by the `Top` level `targetRef`.
```

docs/docs/dev/policies/matching.md
```diff
@@ -0,0 +1,95 @@
...
+        name: default
+      default:
+        backends:
+          - file:
+              format:
+                plain: '{"start_time": "%START_TIME%"}'
+              path: '/tmp/logs.txt'
+```
+
+This policy will target a service `web-frontend`.
```
lahabana, line 93:
```suggestion
This policy will target all proxies that implement the service `web-frontend`.
```

I think we should stop talking a `service` when we're talking about proxies that run a service. It's less confusing

docs/docs/dev/policies/matching.md
```diff
@@ -0,0 +1,95 @@
...
+It instructs Kuma to log traffic **coming** from `web-backend` to `web-frontend`.
```
lahabana, line 94:
```suggestion
It logs traffic **coming** from `web-backend` to `web-frontend`.
```

docs/docs/dev/policies/matching.md
```diff
@@ -0,0 +1,95 @@
...
+It also instructs to log traffic going out of `web-frontend` to **anything** in the `Mesh` named `default`.
```
lahabana, line 95:
```suggestion
It also logs traffic going out of proxies that implement `web-frontend` to **anything** in the `Mesh`.
```

docs/docs/dev/policies/matching.md
```diff
@@ -0,0 +1,95 @@
...
+      default:
+        backends:
+          - file:
+              format:
+                plain: '{"start_time": "%START_TIME%"}'
+              path: '/tmp/logs.txt'
+  from:
+    - targetRef: # from level targetRef
+        kind: Mesh
+        name: default
```
lahabana, line 84: Let's not put `name` in Mesh we don't do XMesh stuff and it's just confusing user atm. Actually adding validation to disable setting it

docs/docs/dev/policies/mesh-trace.md
```diff
@@ -29,6 +29,19 @@ For HTTP you can also manually forward the following headers:
...
+| TargetRef type    | top level | to  | from |
+|-------------------|-----------|-----|------|
+| Mesh              | ✅         | ❌   | ❌    |
+| MeshSubset        | ✅         | ❌   | ❌    |
+| MeshService       | ✅         | ❌   | ❌    |
+| MeshServiceSubset | ✅         | ❌   | ❌    |
+| MeshGatewayRoute  | ✅         | ❌   | ❌    |
+| MeshHTTPRoute     | ❌         | ❌   | ❌    |
+
+[Legend](matching.md#matching-matrix)
```
lahabana, line 43:
```suggestion
If you don't understand this table you should read [matching docs](matching.md#matching-matrix)
```

lahabana (commented):
Good additions! I think your new page will need to be extended with extra info to help users get started with the new policy but that can be done later

docs/docs/dev/policies/mesh-trace.md
slonka: added

docs/docs/dev/policies/matching.md
```diff
@@ -0,0 +1,96 @@
...
+A policy might not support every level
+(e.g. `MeshTrace` only allows top level `targetRef`).
+
+A combination of a policy and the top level `targetRef` can influence which types are available in `to` and `from` levels
+(e.g. `MeshAccessLog` with a top level `targetRef` of `MeshGatewayRoute` can only have `from` level because there is no outbound listener).
+
+If a policy does not support a level then every type will be marked with ❌ in that level
+and the YAML configuration must omit that field from the definition.
+
+# targetRef levels
```
lahabana, line 51:
```suggestion
## targetRef levels
```

Also I think this section should be before the `matrix` one

lahabana (approved):

PR #1039, opened 2022-10-18
docs(deployments/multi-zone): make `curl` examples clearer

michaelbeaumont:


docs/docs/1.8.x/deployments/multi-zone.md
```diff
@@ -350,21 +350,25 @@ spec:
...
+
+To send a request in the same zone, you can rely on Kubernetes DNS and use the usual Kubernetes hostnames and ports:
 
 ```sh
-<kuma-enabled-pod>$ curl http://echo-server:1010
+curl http://echo-server:1010
 ```
 
-To consume the example service across all zones in your Kuma deployment (that is, from endpoints ultimately connecting to the same global control plane), you can run:
+To send a request from any zone, you can [use the generated `kuma.io/service`](../explore/dpp-on-kubernetes.md#tag-generation) and [Kuma DNS](../networking/dns.md#dns):
```
lahabana, line 362:
```suggestion
To send a request to any zone, you can [use the generated `kuma.io/service`](../explore/dpp-on-kubernetes.md#tag-generation) and [Kuma DNS](../networking/dns.md#dns):
```

lahabana: Might be worth saying that by default it will round robin between all zones and you can use localityAwareLB for more specific stuff

lahabana (approved):

PR #1067, opened 2022-10-26
Add sitemap.xml and robots.txt

mheap:
Signed-off-by: Michael Heap <m@michaelheap.com>

Add missing `sitemap.xml` and `robots.txt`

Did you sign your commit? Yes

### DIFF
--- a/app/_plugins/generators/sitemap.rb
+++ b/app/_plugins/generators/sitemap.rb
@@ -1,0 +1,67 @@
+module Sitemap
+  class Generator < Jekyll::Generator # rubocop:disable Metrics/ClassLength
+    priority :lowest
+
+    def generate(site)
+      # What's our latest version?
+      latest = site.data['versions'].reject { |v| v['release'] == 'dev' }.last
+
+      # Grab all pages that contain that version
+      all_pages = []
+      # Build a map of the latest available version of every URL
+      site.pages.each do |page|
+        next if is_versioned(page['url']) and !is_latest(page['url'], latest)
+        all_pages << page
+      end
+
+      # Build a map of the latest available version of every URL
+      site.posts.docs.each do |post|
+       all_pages << {
+        'url' => post.url
+       }
+      end
+
+      # Save the data to generate a sitemap later
+      site.data['sitemap_pages'] = build_sitemap(all_pages)
+    end
+
+    def build_sitemap(pages) # rubocop:disable Metrics/MethodLength
+      # These files should NOT be in the sitemap
+      blocked_from_sitemap = [
+        '/_headers',
+        '/_redirects',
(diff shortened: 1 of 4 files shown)


app/_plugins/generators/sitemap.rb
```diff
@@ -0,0 +1,67 @@
+module Sitemap
+  class Generator < Jekyll::Generator # rubocop:disable Metrics/ClassLength
+    priority :lowest
+
+    def generate(site)
+      # What's our latest version?
+      latest = site.data['versions'].reject { |v| v['release'] == 'dev' }.last
```
lahabana, line 7: versions.yml has lost the `latest` boolean that used to exist. This `latest` boolean is super useful as it enables us to ship the docs for the release during stabilization week without making it indexed by Google and still show the real current one.

In other words as soon as we'd cut the 2.0 release the 2.0 docs would start to exist but `latest` on the website would still be `1.8`.
Does that make sense?

app/robots.txt
```diff
@@ -0,0 +1,5 @@
+User-agent: * 
+Disallow: /latest_version 
+Disallow: /latest_version.html 
+  
+Sitemap: https://kuma.io/sitemap.xml
```
lahabana, line 5: IIRC the robots.txt was pretty simple but there was a bunch a meta on the pages of non latest docs to skip indexing them.

lahabana (commented):
I think latest was useful feature

app/_plugins/generators/sitemap.rb
lahabana: This was discussed live and we decided to do this in the next PR :)

app/_plugins/generators/sitemap.rb
```diff
@@ -0,0 +1,67 @@
...
+        "/install/",
+        "/docs/",
+      ]
+      versioned.each do |v|
+        return true if url.include?(v)
+      end
+      false
+    end
+
+    def is_latest(url, latest)
```
lahabana, line 63: nit: this is more like `is_of_version(url, version)` as the latest param could be any version in practice.

lahabana (approved):
LGTM! There's a question but no reason to block this PR!

app/_plugins/generators/sitemap.rb
mheap: It makes sense - will re-add `latest` in a followup PR

app/robots.txt
mheap: Will add this in a followup

mheap: I've just investigated and the metadata is present on non-latest docs. They're set to `noindex,follow` which is correct

PR #1073, opened 2022-10-26
Generate redirects at build time

mheap:
Generate the redirects file in line with https://github.com/kumahq/kuma-website/blob/584d4f5e0334a951000c3d8215bc3be6fcf7869f/docs/.vuepress/config.js

Did you sign your commit? yes


cloudjumpercat:
@mheap Does this make it so any files that were moved/had an updated filename between 1.8 and 2.0 will have a redirect automatically generated?

mheap:
It doesn't. We'll still need to add those redirects manually

app/_plugins/generators/redirects.rb
```diff
@@ -0,0 +1,129 @@
...
+        "/docs/:version/documentation/deployments/ /docs/:version/introduction/deployments 301",
+        "/latest_version.html /latest_version 301",
+      ].join("\n")
+
+      # Generate redirects for specific versions
+      redirects = "#{redirects}\n\n# Specific version to release\n"
+      active_versions.each do |v|
+        vp = v['version'].split('.').map(&:to_i)
+        idx = 0
+        loop do
```
lahabana, line 38: nit: isn't `for idx in 0.. vp[2] do` a little easier to follow than the loop with a break?

In any case I think a comment explaining that you are generating redirects for all patch versions from `x.y.0` until latest version maybe useful for future readers

lahabana (approved):
Just a nit and I've checked we seem to be at parity with the old vuepress on redirect!

PR #1072, opened 2022-10-26
fix: add back missing MeshTrace doc

lahabana:
It got lost as part of the replatform.
Also improved the docs with more examples and more info.

Signed-off-by: Charly Molter <charly.molter@konghq.com>

PR #1108, opened 2022-10-31
Fix missplaced files after merge

fabianrbz:
Looks like some files were misplaced after we merged the Jekyll migration.
I  put them where they belong and updated their syntax.

@lahabana currently `app/docs/dev/policies/mesh-trace.md` is not listed in the sidebar, where should we place it? I also added a fix to the sidebar's js so that it doesn't error when the current page is not present in the sidebar.

cc @mheap


lahabana:
@fabianrbz meshtrace should be in the sidebar only for `dev` and I can see it: https://kuma.io/docs/dev/ it was merged in #1072 if I'm not wrong

lahabana:
I see for some reason MeshTrace ended up in 1.6 where it really shouldn't exist. Please remove it instead :)

fabianrbz:
looks like it got renamed at some point, I deleted the old one and fixed one more file

lahabana (approved):
LGTM except that dpp-specification shouldn't be part of any docs above 1.6.x (thus it was ok not in `dev`) this is because the file was replaced after 1.6.x by auto-generated docs.

PR #1122, opened 2022-11-02
(docs) Add product variable for Kuma

cloudjumpercat:
Signed-off-by: Diana <75819066+cloudjumpercat@users.noreply.github.com>

- Added a variable for Kuma so that it can be replaced with "Kong Mesh" when the content is shared with the Mesh repo. 
- Replaced "Kuma" (where it made sense) with the variable. I didn't replace it in the generated docs because I knew those would be generated and I didn't replace it in places where it was in code, like `kuma-cp`. 


lahabana:
Also kuma-system (the default namespace) should be replaced by kong-mesh-system everywhere.

cloudjumpercat:
@lahabana Is there any other names, like "kuma-cp" or "kuma-dp" that should also be replaced with a variable, or is it just "kuma-system"?

lahabana:
No just `kuma-system`

jekyll-dev.yml
```diff
@@ -64,3 +64,8 @@ jekyll-generator-single-source:
   layout: 'page'
   multiple_products: false
   base_dest_path: 'docs'
+
+# Product name variables
+base_gateway: Kong Gateway
+mesh_product_name: Kuma
+kic_product_name: Kubernetes Ingress Controller
```
lahabana, line 71: Why are we adding stuff about kic and gateway here?

cloudjumpercat:
> Can we add a linter too as part of this to disallow people to use Kuma directly?

@lahabana We originally did that in the docs.konghq.com repo for our other products, like Kong Gateway, but opted to remove it. I'd recommend not adding a linter here either because sometimes I left "Kuma" as-is because it was specifically referring to the CNCF project and we wouldn't want that replaced with "Kong Mesh".

lahabana (approved):
Proactively approving this. I think there's likely going to be misses but we'll just fix them as we go! 
Please do socialise this in #team-mesh as soon as it's merged as it's something people have to use.

PR #1131, opened 2022-11-03
update general targetRef policy docs

### LINKED ISSUE #1031: Docs for a new policy matching
### Description

We need docs that describe mechanism of a new policy matching

lahabana:
Fix #1031 

### DIFF
--- a/app/docs/dev/policies/targetref.md
+++ b/app/docs/dev/policies/targetref.md
@@ -4,20 +4,21 @@ title: Understanding TargetRef policies
 
 ## What is a policy?
 
-A policy is anything that can modify the configuration of a proxy.
+A policy is a set of configuration that will be used to generate the proxy configuration.
+{{ site.mesh_product_name }} combines policies with dataplane configuration to generate the Envoy configuration of a proxy.
 
 ## What do `targetRef` policies look like?
 
-There are 2 useful parts in a policy:
+There are two parts in a policy:
 
 1. The metadata
 2. The spec
 
-### The metadata
+### Metadata
 
 Metadata identifies the policies by its `name`, `type` and what `mesh` it is part of.
 
-It looks like the following:
+This is how it looks:
 
 {% tabs metadata %}
 {% tab metadata Universal %}
@@ -73,7 +74,7 @@ Policies are namespaced scope and currently the namespace must be the one the co
 {% endtab %}
 {% endtabs %}
 
-### The spec
+### Spec
 
 The spec contains the actual configuration of the policy.
 
@@ -125,7 +126,7 @@ spec:
 {% tip %}
 One of the benefits of `targetRef` policies is that the spec is always the same between Kubernetes and Universal.
 
-This means that converting policies between universal and Kubernetes only means rewriting the metadata.
+This means that converting policies between Universal and Kubernetes only means rewriting the metadata.
 {% endtip %}
 
 #### Writing a `targetRef`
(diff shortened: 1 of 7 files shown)


slonka:
Does this close https://github.com/kumahq/kuma-website/issues/1031 ?

app/docs/dev/policies/targetref.md
```diff
@@ -0,0 +1,333 @@
...
+  targetRef:
+    kind: Mesh
+  default:
+    key: value
+```
+
+{% tip %}
+One of the benefits of `targetRef` policies is that the spec is always the same between Kubernetes and Universal.
+
+This means that converting policies between universal and Kubernetes only means rewriting the metadata.
```
slonka, line 128:
```suggestion
This means that converting policies between Universal and Kubernetes only means rewriting the metadata.
```

app/docs/dev/policies/targetref.md
```diff
@@ -0,0 +1,333 @@
...
+
+`targetRef` is a concept borrowed from [Kubernetes Gateway API](https://gateway-api.sigs.k8s.io/) its usage is fully defined in [MADR 005](https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/005-policy-matching.md).
+Its goal is to select subsets of proxies with maximum flexibility.
+
+It looks like:
+
+```yaml
+targetRef:
+  kind: Mesh | MeshSubset | MeshService | MeshServiceSubset | MeshGatewayRoute
+  name: "my-name" # For kinds MeshService, MeshServiceSubset and MeshGatewayRoute a name can be defined
```
slonka, line 141: Why can't all of them have a name?

app/docs/dev/policies/targetref.md
```diff
@@ -0,0 +1,333 @@
...
+To help users, each policy documentation includes a table indicating which `targetRef` kinds is supported at each level.
+
+This table looks like:
+
+| `targetRef.kind`    | top level | to  | from |
+| ------------------- | --------- | --- | ---- |
+| `Mesh`              | ✅        | ✅  | ❌   |
+| `MeshSubset`        | ✅        | ❌  | ❌   |
+| `MeshService`       | ✅        | ❌  | ✅   |
+| `MeshServiceSubset` | ✅        | ❌  | ❌   |
```
slonka, line 210: We mention MeshGatewayRoute on line [152](https://github.com/kumahq/kuma-website/pull/1131/files#diff-f25d24e526ca4764872874b844778d34817a88616a050f4962aa713fc50e5c08R152) but not here, should we make this consistent?

app/docs/dev/policies/targetref.md
```diff
@@ -0,0 +1,333 @@
...
+
+Here it indicates that the top level can use any targetRef kinds. But in `targetRef.to` only kind `Mesh` can be used and in `targetRef.from` only kind `MeshService`.
+
+
+### Merging configuration
+
+Because for a specific proxy multiple `targetRef` can be used.
+It is necessary to define a policy for merging configuration.
```
slonka, line 218:
```suggestion
It is necessary to define a strategy for merging configurations, 
because a proxy can be targeted by multiple `targetRef`s.
```

app/docs/dev/policies/targetref.md
```diff
@@ -0,0 +1,333 @@
...
+
+We define a total order this way:
```
slonka, line 220:
```suggestion
We define a total order:
```

app/docs/dev/policies/targetref.md
```diff
@@ -0,0 +1,333 @@
...
+
+- Mesh > MeshSubset > MeshService > MeshServiceSubset > MeshGatewayRoute (The more a `targetRef` is specific the higher priority it has)
```
slonka, line 222: I would not use "specific" here because it was used in the S&D explanation.

```suggestion
- Mesh > MeshSubset > MeshService > MeshServiceSubset > MeshGatewayRoute (the more a `targetRef` is focused the higher priority it has)
```

app/docs/dev/policies/targetref.md
```diff
@@ -0,0 +1,333 @@
...
+```
+
+The merge result is:
+```yaml
+conf: 1
+sub:
+  array: []
+  other-array: [5, 6]
+  extra: 2
+```
```
slonka, line 251: Can we have this prefixed with `spec` or some other way link this to the actual policy "world" so that the user is familiar with the constructs?

app/docs/dev/policies/targetref.md
```diff
@@ -0,0 +1,333 @@
+---
+title: Understanding TargetRef policies
+---
+
+## What is a policy?
+
+A policy is anything that can modify the configuration of a proxy.
```
Automaat, line 7: This definition is a little vague for me. If I were unfamiliar with Kuma and it's policies, I wouldn't understand this definition. [Here](https://kuma.io/docs/1.8.x/explore/dpp/#dynamic-configuration-of-the-data-plane-proxy) docs say that policies are used to generate DP configuration. So maybe something like this: 
`A policy is a set of guidelines that will be used to build proxy configuration. Kuma combines policies and Dataplane entities to build configuration for proxy`

app/docs/dev/policies/targetref.md
```diff
@@ -0,0 +1,333 @@
...
+
+## What do `targetRef` policies look like?
+
+There are 2 useful parts in a policy:
```
Automaat, line 11: `useful parts` suggests that these are not mandatory fields. Are they?

app/docs/dev/policies/targetref.md
```diff
@@ -0,0 +1,333 @@
...
+conf: 1
+sub:
+  array: []
+  other-array: [5, 6]
+  extra: 2
+```
+
+### Examples
+
+#### Applying as a global default
```
slonka, line 255:
```suggestion
#### Applying a global default
```

slonka (commented):
Left a bunch of comments, some of them just suggestions, feel free to not apply them if you don't strongly agree. On a more general thought I think we could benefit from a real world example with overwrites.

app/docs/dev/policies/targetref.md
lahabana: Cause for Mesh/MeshSubset you don't have a choice right now (it must be the one that the policy is in) In the future we might relax this but not now

Automaat (approved):

Issue #635, opened 2021-12-23
Document timeouts and connection limits for connections outside of the Mesh


jakubdyszkiewicz:
### Description

Since we don't let users configure those values, we should at least document them.

https://github.com/kumahq/kuma/issues/2905
https://github.com/kumahq/kuma/issues/3614

PR #1139, opened 2022-11-04
docs: timeouts and thresholds for outside mesh connections

lobkovilya:

Fix https://github.com/kumahq/kuma-website/issues/635


lahabana (changes requested):
I think it makes more sense to have this its own page in `networking` and then link it from timeout and circuit-breaker

lahabana:
Let's delay it to after release

lobkovilya:
Why do you think it should be a common page for timeouts and circuit-breaker? For me, it seems okay to have different pages, because:
* There are no common parts/concepts to share.
* If you troubleshoot timeouts you most likely end up on a timeout page, trying to find a way to change timeouts for outside traffic. The same applies to a circuit breaker.

lahabana:
I think my point is that there should be a page specific to non mesh traffic where we explain timeouts etc rather than have these in the docs of each policy

lobkovilya:
But this new page specific to non mesh traffic will have a small intro paragraph and a section for each policy. Why do we want to collect all these sections on one page?

lahabana:
I think it's important to put yourself in the user's shoes. Something is odd with their traffic that's not part of the mesh, where do you think they'll be looking for info?

lobkovilya:
I agree, there are 2 possible ways users can navigate the docs:
* Requests are terminated due to "timeout exceeded" –> go to the Timeouts page. Counter `upstream_cx_overflow` is increasing –> go to the Circuit Breaker page. 
* User observes issues with non-mesh traffic –> go to "Networking/Outside mesh traffic".

Ideally, our documentation is helpful in both of these patterns. We can create a page in Networking, describing `passthrough` cluster, and how to disable it and we link here the relevant "Outside mesh traffic" sections from the policy pages. In that case, we keep timeout-related information in `Timeouts`, information about circuit breaker thresholds in `Circuit Breaker`, etc. WDYT?

lahabana:
Sounds good

app/_data/docs_nav_kuma_dev.yml
```diff
@@ -268,6 +268,8 @@ items:
         items:
           - text: Disabling IPv6
             url: "/networking/ipv6/#disabling-ipv6"
+      - text: Non-mesh traffic
+        url: /networking/non-mesh-traffic/
```
lahabana, line 272: Doesn't that goes under `transparent-proxy` ?

app/docs/dev/networking/non-mesh-traffic.md
```diff
@@ -0,0 +1,202 @@
...
+  outbound:
+    passthrough: false
+```
+{% endtab %}
+{% endtabs %}
+
+When `networking.outbound.passthrough` is `false`, no traffic to any non-mesh resource can leave the Mesh.
+
+{% tip %}
+Before turning this feature on, double-check Envoy stats that no traffic is flowing through `pass_through` cluster.
```
lahabana, line 38:
```suggestion
Before turning this feature off, double-check Envoy stats that no traffic is flowing through `pass_through` cluster. 
```
It's setting `passthrough=false` that's dangerous right?

app/docs/dev/networking/non-mesh-traffic.md
```diff
@@ -0,0 +1,202 @@
+---
+title: Non-mesh traffic
+---
+
+In its default setup, {{site.mesh_product_name}} allows any non-mesh traffic to pass Envoy without applying any policy.
```
lahabana, line 5: Should we split this in 2? Incoming and Outgoing traffic?

I feel like it would be useful to cover incoming traffic too? Questions about this come up now and then.
Say something about impossible with mtls, link to permissive. Say that without transparent proxy people can bypass if you don't protect...

app/_data/docs_nav_kuma_dev.yml
lobkovilya: "non-mesh-traffic.md" is a separate page that contains all the information about the outside mesh traffic. So it's not under "transparent-proxy"

lobkovilya:
@Mergifyio update

app/docs/dev/networking/non-mesh-traffic.md
```diff
@@ -0,0 +1,215 @@
...
+{% endtabs %}
+
+When `networking.outbound.passthrough` is `false`, no traffic to any non-mesh resource can leave the Mesh.
+
+{% tip %}
+Before turning this feature on, double-check Envoy stats that no traffic is flowing through `pass_through` cluster. 
+Otherwise, you will block the traffic which may cause the instability of the system.
+{% endtip %}
+
+### Policies don't apply to the non-mesh traffic
```
lahabana, line 55:
```suggestion
### Policies don't apply to non-mesh traffic
```

app/docs/dev/networking/non-mesh-traffic.md
```diff
@@ -0,0 +1,215 @@
...
+
+If you need to change configuration for the non-mesh traffic
```
lahabana, line 57:
```suggestion
If you need to change configuration for non-mesh traffic 
```

lahabana (approved):

app/docs/dev/networking/non-mesh-traffic.md
lahabana:
```suggestion
Before setting `networking.outbound.passthrough` to `false`, double-check Envoy stats that no traffic is flowing through `pass_through` cluster. 
```

lahabana (approved):

PR #1144, opened 2022-11-07
docs(MeshGateway): try to make multi-zone clearer

### LINKED ISSUE #912: Clarify in the docs in multi-zone what needs to be done in global and what needs to be done in zone
### What happened?

It's just unclear atm. Maybe this should be a table of things?

michaelbeaumont:
Closes #912


slonka (approved):

slonka:
Maybe someone who had problems with this could review, do you remember who was that? (like who caused us to create the issue)

michaelbeaumont:
@lahabana do you remember? I think it's happened multiple times to be honest. I think it's usually a `MeshGatewayInstance` being created in global.

I can't think of anything to do except call it out where we introduce the resources. `MeshGatewayInstance` in particular doesn't have its own "policy page" so this is the only place.

I thought we had validation in the webhook for tis actually but I don't see it in fact. That's something to check/add as well.

PR #1149, opened 2022-11-09
refactor: include preview.sh script into installer.sh

### LINKED ISSUE #960: Release "dev" shows kumactl 1.5.0 in installation instructions
### What happened?

image

The version should probably be the current preview version.

kleinfreund:
Incorporates the logic from the preview.sh script into the installer.sh script.

Changes the preview.sh script to use the installer.sh script directly with `VERSION=preview`.

Sets the latest version for the dev docs to `preview` (was `2.0.0`).

Fixes broken links to the Kubernetes documentation.

Resolves #960.


app/_includes/snippets/install_os.md
```diff
@@ -11,7 +11,7 @@ Finally, you can follow the [Quickstart](#quickstart) to take it from here and c
 Run the following script to automatically detect the operating system and download Kuma:
 
 <div class="language-sh">
-<pre class="no-line-numbers"><code>curl -L https://kuma.io/installer.sh | VERSION={{ page.latest_version }} sh -</code></pre>
+<pre class="no-line-numbers"><code>curl -L https://kuma.io/installer.sh | VERSION={{ include.installer_version }} sh -</code></pre>
```
lahabana, line 14: I don't think these are needed anymore. You can use page.latestVersion everywhere this is pulled from: 

https://github.com/kumahq/kuma-website/blob/990729fda76e3258c015d6461b712428afe72361/app/_data/versions.yml#L30

If you update this to `preview` it should work

app/installer.sh
```diff
@@ -26,6 +26,11 @@ DIR="$( cd "$( dirname "$0" )" >/dev/null 2>&1 && pwd )"
 : "${LATEST_VERSION:=https://kuma.io/latest_version}"
 : "${REPO_PREFIX:=kuma}"
 : "${CTL_NAME:=kumactl}"
+: "${BRANCH:=master}"
+
+if [ "$REPO_PREFIX" == "kong-mesh" ]; then
+  PRODUCT_NAME="KONG MESH"
+fi
```
lahabana, line 33: For kong-mesh here's how we invoke this script: 

https://github.com/Kong/docs.konghq.com/blob/main/app/mesh/installer.sh#L21-L25

So what should be done is set in docs.konghq.com an extra REPO parameter instead of doing the special casing here

app/installer.sh
```diff
@@ -159,3 +168,42 @@ else
   printf "ERROR\tUnable to download %s\n" "$TARGET_NAME"
   exit 1
 fi
+
+# Sets `VERSION` to the git tag representing the latest preview version.
+set_preview_version() {
+  if [[ ! -x `which gh` ]]; then
+    echo "You must have github's gh client installed to install a preview version"
+    exit 1
+  fi
```
lahabana, line 177: You've lost: 

```sh
IFS=/ read -r owner repo << EOF
${REPO}
EOF
```

app/installer.sh
lahabana: So you'd add after line 29:

```sh
: "${REPO:=kumahq/kuma}"
```

app/installer.sh
```diff
@@ -159,3 +168,42 @@ else
...
+  }
+  '`
+
+  echo "Getting release $VERSION"
+
+  if [ "$REPO_PREFIX" == "kong-mesh" ]; then
+    PREVIEW_COMMIT=${PREVIEW_COMMIT:0:7}
+  else
+    PREVIEW_COMMIT=${PREVIEW_COMMIT:0:9}
+  fi
```
lahabana, line 206: I did it this way because the short commit hash isn't constant (we should use `git rev-parse --short=9` (we could potentially fix the build instead :) ).

Let's fix this in Kuma and get rid of the if block

lahabana (commented):
Some nits but it's in a good path

app/_includes/snippets/install_os.md
kleinfreund: This would be incorrect for any use of `page.latest_version` that isn’t related to the installer script and there are quite a few of those:

```
- **kuma-cp**: at `docker.io/kumahq/kuma-cp:{{ page.latest_version }}`
```

```
You will then find the `kumactl` executable in the `kuma-{{ page.latest_version }}/bin` folder.
```

```
- <a href="https://download.konghq.com/mesh-alpine/kuma-{{ page.latest_version }}-centos-amd64.tar.gz">CentOS</a>
```

So while it’s definitely still incorrect that these refer to 2.0.0, I think it’s better than referring to `preview` which will definitely fail for all of these cases.

app/installer.sh
kleinfreund: That’s back in now.

app/installer.sh
kleinfreund: For now I replaced this with a sh-compatible version, but that would be nice to sort out in Kuma/Kong Mesh.

app/installer.sh
kleinfreund: I’ve added support for this and made the preview.sh script use installer.sh script under the hood for people that use it (muscle memory).

Do you recall another place where we might use `REPO_PREFIX` instead of `REPO` like that? I’ll take care of the Kong Mesh docs one after this.

app/_includes/snippets/install_os.md
lahabana: Having preview here for the sake of simplicity is better imo. What you are doing is also making single source docs complicated.
We can always push the docker tag on preview builds it's not that complicated

app/docs/dev/installation/docker.md
```diff
@@ -114,7 +114,7 @@ mtls:
 You can run the following script to automatically detect the operating system and download {{site.mesh_product_name}}:
 
 ```sh
-curl -L https://kuma.io/installer.sh | sh -
+curl -L https://kuma.io/installer.sh | VERSION=preview sh -
```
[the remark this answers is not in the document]
lahabana: Still correct

app/installer.sh
lahabana: I'd keep PRODUCT_NAME and not have anything specific for kong-mesh in here

app/_includes/snippets/install_os.md
kleinfreund: I’ve changed this again and use the preview version set in versions.yml now.

app/installer.sh
kleinfreund: I’ve added `PRODUCT_NAME` again and also added in preview.sh again.

I’ve also changed the detection of `REPO_PREFIX` to strip out anything up to and including the slash from `REPO`.

The remaining reference to Kong Mesh is the different preview version hash which we’re fixing separatetely.

app/installer.sh
```diff
@@ -23,10 +23,13 @@ DIR="$( cd "$( dirname "$0" )" >/dev/null 2>&1 && pwd )"
 : "${VERSION:=}"
 : "${ARCH:=}"
 : "${PRODUCT_NAME:=Kuma}"
+: "${REPO:=kumahq/kuma}"
+: "${BRANCH:=master}"
 : "${LATEST_VERSION:=https://kuma.io/latest_version}"
-: "${REPO_PREFIX:=kuma}"
 : "${CTL_NAME:=kumactl}"
 
+REPO_PREFIX=$(echo $REPO | sed 's/.*\///')
```
lahabana, line 31:
```suggestion
REPO_PREFIX=$(echo $REPO | sed 's|.*/||')
```

I find this a bit more readable but up to you to ignore this

app/installer.sh
```diff
@@ -107,6 +110,53 @@ if [ "$OS" = "Linux" ]; then
...
+# Sets `VERSION` to the git tag representing the latest preview version.
+set_preview_version() {
+  if [ ! -x `which gh` ]; then
+    echo "You must have github's gh client installed to install a preview version"
+    exit 1
+  fi
+
+  IFS=/ read -r owner repo << EOF
+${REPO}
+EOF
```
lahabana, line 122: I guess: REPO_PREFIX is the repo here.

app/installer.sh
```diff
@@ -107,6 +110,53 @@ if [ "$OS" = "Linux" ]; then
...
+    }
+  }
+}
+'`
+
+  if [ "$REPO" = "kong/kong-mesh" ]; then
+    PREVIEW_COMMIT=$(echo $PREVIEW_COMMIT | cut -c -7)
+  else
+    PREVIEW_COMMIT=$(echo $PREVIEW_COMMIT | cut -c -9)
+  fi
```
lahabana, line 150: Please do the change to always use 9 I'll open the PR to fix the conf today

lahabana (approved):

app/installer.sh
lahabana: It's here: https://github.com/kumahq/kuma/pull/5296

app/installer.sh
kleinfreund: I like this better as well, changing it.

app/installer.sh
kleinfreund: I’m not following. What do you mean by this?

app/installer.sh
kleinfreund: Done.

app/installer.sh
lahabana: When you do: `REPO_PREFIX=$(echo $REPO | sed 's|.*/||')` you end up with REPO_PREFIX that is equal to you `repo` variable extracted here. Maybe it's good enough to have just once

kleinfreund: Ah, now I get it. Yes, indeed. I’ve replaced this `read` usage altogether now and construct a new `REPO_ORG` variable in the same `sed` style as `REPO_PREFIX`.

kleinfreund:
**Comment**: Waiting with merging this until we’ve published the version Kong Mesh preview version with a 9 character hash.

kleinfreund:
**Comment**: The first Kong Mesh preview version with a 9 character hash was published (e.g. https://download.konghq.com/mesh-alpine/kong-mesh-0.0.0-preview.v0d9c23609-ubuntu-amd64.tar.gz) so I’m merging this.

michaelbeaumont:
@kleinfreund next time can you open a new PR to fix things like the service account links? It has nothing to do with the `preview.sh` script

PR #1170, opened 2022-12-13
fix(meshaccesslog): subset bad example

lahabana:
We were using MeshService with tags. Replaced with MeshServiceSubset


slonka (approved):

slonka:
DCO missing

lahabana:
Weird I fixed the DCO I don't understand why it's not taking it into account

PR #1203, opened 2023-01-23
Single source generation for `dev`

fabianrbz:
Single source `dev`.

* Bump `jekyll-generator-single-source` which adds support for `dev` as a version.
* Add missing pages to the nav file under `unlisted` so they get generated but not rendered in the sidenav.


cloudjumpercat (approved):
I tested this by using {% if_version eq:2.1.x %} on one page and that content only displayed for the dev version of the docs, so everything looks good to me!

cloudjumpercat:
@fabianrbz Okay, I just tested all of the following in the single source files and everything rendered correctly:

```
{% if_version eq:2.1.x %}
Should display for 2.1.x eq
{% endif_version %}

{% if_version gte:2.1.x %}
Should display for 2.1.x gte
{% endif_version %}

{% if_version lte:2.1.x %}
Should display for 2.1.x lte (and probably also versions lte than 2.1.x)
{% endif_version %}

{% if_version eq:1.8.7 %}
Should NOT display for 2.1.x (it's eq 1.8.7)
{% endif_version %}

{% if_version lte:1.8.7 %}
Should NOT display for 2.1.x (it's lte 1.8.7)
{% endif_version %}

{% if_version gte:1.8.7 %}
SHOULD display for 2.1.x (it's gte 1.8.7)
{% endif_version %}
```

app/_data/docs_nav_kuma_dev.yml
```diff
@@ -805,3 +807,150 @@ items:
...
+  - url: /generated/cmd/kumactl/kumactl_get_zoneegresses/
+  - url: /generated/cmd/kumactl/kumactl_get_zones/
+  - url: /generated/cmd/kumactl/kumactl_uninstall_ebpf/
+  - url: /generated/cmd/kumactl/kumactl_uninstall_transparent-proxy/
+  - url: /generated/cmd/kumactl/kumactl_install_gateway_kong/
+  - url: /generated/cmd/kumactl/kumactl_install_gateway_kong-enterprise/
+  - url: /generated/cmd/kumactl/kumactl_config_control-planes_add/
+  - url: /generated/cmd/kumactl/kumactl_config_control-planes_list/
+  - url: /generated/cmd/kumactl/kumactl_config_control-planes_remove/
+  - url: /generated/cmd/kumactl/kumactl_config_control-planes_switch/
```
slonka, line 956: do we need to manually update this or is it generated? if it's not generated can't we just use `/generated/*` or something similar so we don't have to update it?

lahabana:
@fabianrbz one thing I see is that generated versions are single sources too which is not possible.
The generated docs are coming straight from Kuma (see [this action](https://github.com/kumahq/kuma-website/blob/master/.github/workflows/sync-generated.yml)) so we should have one folder per version and have the sidebar reference the matching version.

Also this can be done outside of this PR but it would be good as @slonka mentionned to have these part of the sidebar auto-generated.

app/_data/docs_nav_kuma_dev.yml
```diff
@@ -805,3 +807,150 @@ items:
             url: "/contribute/introduction/#gui-code"
           - text: Testing unreleased versions
             url: "/contribute/introduction/#testing-unreleased-versions"
+unlisted:
+  - url: /
+  - url: /documentation/dependencies/
+  - url: /documentation/introduction/
```
lahabana, line 813: I think this page can be removed (it was likely just inherited from the past)

app/_data/docs_nav_kuma_dev.yml
```diff
@@ -805,3 +807,150 @@ items:
             url: "/contribute/introduction/#gui-code"
           - text: Testing unreleased versions
             url: "/contribute/introduction/#testing-unreleased-versions"
+unlisted:
+  - url: /
+  - url: /documentation/dependencies/
```
lahabana, line 812: I think this page can be removed (it was likely just inherited from the past)

app/_data/docs_nav_kuma_dev.yml
```diff
@@ -805,3 +807,150 @@ items:
...
+  - url: /documentation/introduction/
+  - url: /policies/matching/
```
lahabana, line 814: This page was just forgotten to be removed when it was replaced by `/policies/targetRef`

app/_data/docs_nav_kuma_dev.yml
fabianrbz: I manually added them, but that was before realizing that we don't need to single-source the generated files. Hence, we don't need to add the whole list, just the ones we want to see rendered on the sidenav.

app/_data/docs_nav_kuma_dev.yml
fabianrbz: Loos like this one is sill referenced [here](https://github.com/kumahq/kuma-website/blob/master/app/docs/dev/documentation/configuration.md?plain=1#L101)

lahabana: Can we just rewrite this line to:

When {{site.mesh_product_name}} (`kuma-cp`) is up and running it needs to store its state on Universal it's Postgres, for Kubernetes it's leveraging Kubernetes custom resource definitions.

Please :)

WRITING-DOCUMENTATION.md
```diff
@@ -22,15 +31,9 @@ Ask a maintainer to get write access.
 
 ## Cutting a new release
 
-To cut the dev release copy paste the `dev` folder and rename it to the correct version:
+To cut the dev release, create a duplicate of the [docs_nav_kuma_dev.yml](app/_data/docs_nav_kuma_dev.yml) file and then rename one of the files to "docs_nav_kuma_[version].yml". Update the `release: dev` metadata in the new release file with the release version.
```
lahabana, line 34: @Automaat FYI this is what you'll need to do to cut 2.1

lahabana (approved):
Ship it! Super happy on getting this in!

WRITING-DOCUMENTATION.md
```diff
@@ -7,12 +7,21 @@ as you write your documentation.
...
+A single sourced folder in [app/_src](app/_src) is used for each version of Kuma. We use a Jekyll plugin to dynamically generate pages from a single source file.
+
+For the future non-patch versions of Kuma, changes can be made to the [docs_nav_kuma_dev.yml](app/_data/docs_nav_kuma_dev.yml) file. 
 
 ## Writing docs for a new feature
 
-If you are writing docs for a new feature you'll want to add it in the [dev](app/docs/dev) folder.
+If you are writing docs for a new feature you'll want to add it in the [src](app/_src) folder.
+
+Since content is single sourced, you must use [conditional rendering](link_to_cond_rendering) to ensure that the new feature content only displays for that version. For example:
```
cloudjumpercat, line 18:
```suggestion
Since content is single sourced, you must use [conditional rendering](https://docs.konghq.com/contributing/conditional-rendering/) to ensure that the new feature content only displays for that version. For example:
```

lahabana:
@fabianrbz @cloudjumpercat looks like it's ready can I merge to unblock other PRs?

WRITING-DOCUMENTATION.md
```diff
@@ -7,12 +7,21 @@ as you write your documentation.
...
 
 ## Writing docs for a new feature
 
-If you are writing docs for a new feature you'll want to add it in the [dev](app/docs/dev) folder.
+If you are writing docs for a new feature you'll want to add it in the [src](app/_src) folder.
+
+Since content is single sourced, you must use [conditional rendering](https://docs.konghq.com/contributing/conditional-rendering/) to ensure that the new feature content only displays for that version. For example:
+
+```
+{% if_version eq:2.1.x %}
```
lahabana, line 21: Should this be `gt:2.1.x` ?

cloudjumpercat: `eq` is correct. That means the version has to be equal to 2.1.x for the content to display. If we used `gte`, it would display for 2.1.x and later versions.

lahabana: I believe it makes sense to use `gt` as we'd want the new feature also available in future version's docs, we usually don't put docs for just one version (except for behaviour that would be fixed later but that's fairly rare)?

Maybe just enumerating the options (linking to the external docs for syntax is enough) and let the writer pick what's best is what we should do here.

WRITING-DOCUMENTATION.md
```diff
@@ -7,12 +7,21 @@ as you write your documentation.
...
+A single sourced folder in [app/_src](app/_src) is used for each version of Kuma. We use a Jekyll plugin to dynamically generate pages from a single source file.
+
+For the future non-patch versions of Kuma, changes can be made to the [docs_nav_kuma_dev.yml](app/_data/docs_nav_kuma_dev.yml) file. 
 
 ## Writing docs for a new feature
 
-If you are writing docs for a new feature you'll want to add it in the [dev](app/docs/dev) folder.
+If you are writing docs for a new feature you'll want to add it in the [src](app/_src) folder.
+
+Since content is single sourced, you must use [conditional rendering](https://docs.konghq.com/contributing/conditional-rendering/) to ensure that the new feature content only displays for that version. For example:
```
slonka, line 18: How do we make sure a page is only displayed for version X? Is not including it in the `docs_nav_...` file enough? Like if someone mistakenly links to it somewhere else will it still work?

lahabana: It will but then people would introduce a bug.

Issue #1198, opened 2023-01-16
MeshFaultInjection docs


lahabana:
### Description

Add docs for this new policy

PR #1211, opened 2023-01-25
docs(policies): added docs for MeshFaultInjection

lukidzi:
Added docs for MeshFaultInjection

Fix: #1198


app/_src/policies/meshfaultinjection.md
```diff
@@ -0,0 +1,253 @@
...
+---
+
+With the MeshFaultInjection policy you can easily test your microservices against resiliency.
+
+{% warning %}
+This policy uses a new policy matching algorithm and is in beta state. It should not be combined with [FaultInjection](/docs/{{ page.version }}/policies/fault-injection).
+{% endwarning %}
+
+`MeshFaultInjection` policy is available only for L7 HTTP traffic,
+therefore `kuma.io/protocol` is a mandatory tag to properly propagate configuration.
```
lobkovilya, line 12: We don't validate this anymore, right? Maybe we shouldn't mention it

app/_src/policies/meshfaultinjection.md
```diff
@@ -0,0 +1,253 @@
...
+| `Mesh`              | ✅        | ❌  | ✅   |
+| `MeshSubset`        | ✅        | ❌  | ✅   |
+| `MeshService`       | ✅        | ❌  | ✅   |
+| `MeshServiceSubset` | ✅        | ❌  | ✅   |
+
+To learn more about the information in this table, see the [matching docs](/docs/{{ page.version }}/policies/targetref).
+
+## Configuration
+
+`MeshFaultInjection` allows configuring a list of HTTP faults. They execute in the same order as they were defined.
```
lobkovilya, line 27: Maybe we should mention that every next item in `http` array has a percentage that depends on previous items. We can show that:
```yaml
http:
  - abort:
      httpStatus: 500
      percentage: 70
  - abort:
      httpStatus: 503
      percentage: 50
```
503 will appear in `0.3 * 0.5 * 100` cases because each item is a filter in the filter chain

app/_src/policies/meshfaultinjection.md
```diff
@@ -0,0 +1,263 @@
...
+       percentage: 50
+   - abort:
+       httpStatus: 500
+       percentage: 10
+   - delay:
+       value: 5s
+       percentage: 5
+```
+
+It's worth mentioning that percentage of the next filter depends on the percentage of previous ones.
```
Automaat, line 46: Is it because of how envoy works?

Automaat (approved):

app/_src/policies/meshfaultinjection.md
lukidzi: Yes, that is because filters are applied sequentially, so Filter1 -> Filter2 -> ... -> FilterN and they have the same header match

Issue #1175, opened 2022-12-16
MeshCircuitBreaker docs


slonka:
### Description

docs for - https://github.com/kumahq/kuma/pull/5493

PR #1214, opened 2023-01-26
docs: add MeshCircuitBreaker docs

bartsmykla:
Add MeshCircuitBreaker policy docs

Closes: https://github.com/kumahq/kuma-website/issues/1175

Rendered: https://deploy-preview-1214--kuma.netlify.app/docs/dev/policies/meshcircuitbreaker/

### DIFF
--- a/app/_data/docs_nav_kuma_dev.yml
+++ b/app/_data/docs_nav_kuma_dev.yml
@@ -688,6 +688,8 @@ items:
         items:
           - text: TargetRef support matrix
             url: "/policies/meshcircuitbreaker/#targetref-support-matrix"
+          - text: Configuration
+            url: "/policies/meshcircuitbreaker/#configuration"
           - text: Examples
             url: "/policies/meshcircuitbreaker/#examples"
           - text: Connection limits configuration
--- a/app/_data/docs_nav_kuma_dev.yml
+++ b/app/_data/docs_nav_kuma_dev.yml
@@ -692,10 +692,6 @@ items:
             url: "/policies/meshcircuitbreaker/#configuration"
           - text: Examples
             url: "/policies/meshcircuitbreaker/#examples"
-          - text: Connection limits configuration
-            url: "/policies/meshcircuitbreaker/#connection-limits-configuration"
-          - text: Outlier detection configuration
-            url: "/policies/meshcircuitbreaker/#outlier-detection-configuration"
   - title: Reference docs
     group: true
     items:
--- a/app/_data/docs_nav_kuma_dev.yml
+++ b/app/_data/docs_nav_kuma_dev.yml
@@ -683,6 +683,17 @@ items:
             url: "/policies/meshproxypatch/#examples"
           - text: Merging
             url: "/policies/meshproxypatch/#merging"
+      - text: MeshCircuitBreaker (Beta)
+        url: /policies/meshcircuitbreaker/
+        items:
+          - text: TargetRef support matrix
+            url: "/policies/meshcircuitbreaker/#targetref-support-matrix"
+          - text: Examples
+            url: "/policies/meshcircuitbreaker/#examples"
+          - text: Connection limits configuration
+            url: "/policies/meshcircuitbreaker/#connection-limits-configuration"
+          - text: Outlier detection configuration
+            url: "/policies/meshcircuitbreaker/#outlier-detection-configuration"
   - title: Reference docs
     group: true
     items:
--- a/app/_src/policies/meshcircuitbreaker.md
+++ b/app/_src/policies/meshcircuitbreaker.md
@@ -31,7 +31,7 @@ proxy will further extend the [baseEjectionTime](#outlier-detection) multiplied 
 the 4th ejection will be lasting for a period of time of `4 * baseEjectionTime`.
 
 This policy provides **passive** checks.
-If you want to configure **active** checks, please utilize the [MeshHealthCheck](/docs/{{ page.version }}/policies/meshhealthcheckcz)
+If you want to configure **active** checks, please utilize the [MeshHealthCheck](/docs/{{ page.version }}/policies/meshhealthcheck)
 policy.
 Data plane proxies with **passive** checks won't explicitly send requests to other data plane proxies to determine if
 target proxies are healthy or not.
--- a/app/_src/policies/meshcircuitbreaker.md
+++ b/app/_src/policies/meshcircuitbreaker.md
@@ -1,0 +1,303 @@
+---
+title: MeshCircuitBreaker (beta)
+---
+
+{% warning %}
+This policy uses new policy matching algorithm and is in beta state, it should not be mixed
+with [CircuitBreaker](../circuit-breaker).
+{% endwarning %}
+
+This policy will look for errors in the live traffic being exchanged between our data plane proxies and it will mark a
+data proxy as an unhealthy if certain conditions are met and - by doing so - making sure that no additional traffic can
+reach an unhealthy data plane proxy until it is healthy again.
+
+Circuit breakers - unlike active [MeshHealthChecks](/docs/{{ page.version }}/policies/meshhealthcheck/) - do not send
+additional traffic to our data plane proxies but they rather inspect the existing service traffic. They are also
+commonly used to prevent cascading failures in our services.
+
+{% tip %}
+Like a real-world circuit breaker when the circuit is **closed** then traffic between a source and destination data
+plane proxy is allowed to freely flow through it, and when it is **open** then the traffic is interrupted.
+{% endtip %}
+
+The conditions that determine when a circuit breaker is closed or open are being configured on connection limits or
+outlier detection basis. For outlier detection to open circuit breaker you can configure what we call "detectors".
+This policy provides 5 different types of detectors, and they are triggered on some deviations in the upstream service
+behavior. All detectors could coexist on the same outbound interface.
+
+Once one of the detectors has been triggered the corresponding data plane proxy is ejected from the set of the load
+balancer for a period equal to [baseEjectionTime](#outlier-detection-configuration). Every further ejection of the same data plane
+proxy will further extend the [baseEjectionTime](#outlier-detection-configuration) multiplied by the number of ejections: for example
+the 4th ejection will be lasting for a period of time of `4 * baseEjectionTime`.
+
+This policy provides **passive** checks.
+If you want to configure **active** checks, please utilize the [MeshHealthCheck](/docs/{{ page.version }}/policies/meshhealthcheckcz)
+policy.
+Data plane proxies with **passive** checks won't explicitly send requests to other data plane proxies to determine if
+target proxies are healthy or not.
+
+## TargetRef support matrix
+
+| TargetRef type    | top level | to  | from |
+|-------------------|-----------|-----|------|
+| Mesh              | ✅         | ✅   | ✅    |
+| MeshSubset        | ✅         | ❌   | ❌    |
+| MeshService       | ✅         | ✅   | ❌    |
+| MeshServiceSubset | ✅         | ❌   | ❌    |
+| MeshGatewayRoute  | ❌         | ❌   | ❌    |
+
+To learn more about the information in this table, see the [matching docs](/docs/{{ page.version }}/policies/targetref).
+
+### Examples
+
+#### Basic circuit breaker for outbound traffic from web, to backend service
+
+{% tabs usage useUrlFragment=false %}
+{% tab usage Kubernetes %}
+
+```yaml
+apiVersion: kuma.io/v1alpha1
+kind: MeshCircuitBreaker
+metadata:
+  name: web-to-backend-circuit-breaker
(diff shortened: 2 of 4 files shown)


app/_src/policies/meshcircuitbreaker.md
```diff
@@ -0,0 +1,303 @@
...
+---
+
+{% warning %}
+This policy uses new policy matching algorithm and is in beta state, it should not be mixed
+with [CircuitBreaker](../circuit-breaker).
+{% endwarning %}
+
+This policy will look for errors in the live traffic being exchanged between our data plane proxies and it will mark a
+data proxy as an unhealthy if certain conditions are met and - by doing so - making sure that no additional traffic can
+reach an unhealthy data plane proxy until it is healthy again.
```
jakubdyszkiewicz, line 12: how about shorter sentences? Something like
```
This policy will look for errors in the live traffic being exchanged between our data plane proxies.
It will mark a data proxy as unhealthy if certain conditions are met.
The policy will ensure that no additional traffic can reach an unhealthy data plane proxy until it is healthy again.
```

app/_src/policies/meshcircuitbreaker.md
```diff
@@ -0,0 +1,303 @@
...
+
+Circuit breakers - unlike active [MeshHealthChecks](/docs/{{ page.version }}/policies/meshhealthcheck/) - do not send
+additional traffic to our data plane proxies but they rather inspect the existing service traffic. They are also
+commonly used to prevent cascading failures in our services.
```
jakubdyszkiewicz, line 16: nit: `our services` sounds a bit weird. Maybe `in the services` or just `cascading failures`?

app/_src/policies/meshcircuitbreaker.md
```diff
@@ -0,0 +1,303 @@
...
+
+{% tip %}
+Like a real-world circuit breaker when the circuit is **closed** then traffic between a source and destination data
+plane proxy is allowed to freely flow through it, and when it is **open** then the traffic is interrupted.
```
jakubdyszkiewicz, line 20: `, and when` -> `. When`

app/_src/policies/meshcircuitbreaker.md
```diff
@@ -0,0 +1,303 @@
...
+
+We will apply the configuration with `kumactl apply -f [..]` or via the [HTTP API](/docs/{{ page.version }}/reference/http-api).
+{% endtab %}
+{% endtabs %}
+
+## Connection limits configuration
+
+- **`maxConnections`** - (optional) The maximum number of connections allowed to be made to the upstream cluster. If not
+  specified then equal to "1024".
+- **`maxConnectionPools`** - (optional) The maximum number of connection pools per cluster that are concurrently
```
jakubdyszkiewicz, line 197: not super clear that it's Envoy cluster. Btw. what does it mean that something has multiple connection pools? I thought that there is one connection pool for destination

app/_src/policies/meshcircuitbreaker.md
```diff
@@ -0,0 +1,303 @@
...
+## Outlier detection configuration
+
+- **`disabled`** - (optional) When set to true, outlierDetection configuration won't take any effect.
+- **`interval`** - (optional) The time interval between ejection analysis sweeps. This can result in both new ejections
+  and hosts being returned to service.
+- **`baseEjectionTime`** - (optional) The base time that a host is ejected for. The real time is equal to the base time
+  multiplied by the number of times the host has been ejected.
+- **`maxEjectionPercent`** - (optional) The maximum % of an upstream cluster that can be ejected due to outlier
+  detection. Defaults to 10% but will eject at least one host regardless of the value.
+- **`splitExternalAndLocalErrors`** - (optional) Determines whether to distinguish local origin failures from external
```
slonka, line 216: what are the types of local failures vs external errors?

app/_src/policies/meshcircuitbreaker.md
```diff
@@ -0,0 +1,303 @@
...
+additional traffic to our data plane proxies but they rather inspect the existing service traffic. They are also
+commonly used to prevent cascading failures in our services.
+
+{% tip %}
+Like a real-world circuit breaker when the circuit is **closed** then traffic between a source and destination data
+plane proxy is allowed to freely flow through it, and when it is **open** then the traffic is interrupted.
+{% endtip %}
+
+The conditions that determine when a circuit breaker is closed or open are being configured on connection limits or
+outlier detection basis. For outlier detection to open circuit breaker you can configure what we call "detectors".
```
slonka, line 24: should we go into more details about `outlier detection`?

app/_src/policies/meshcircuitbreaker.md
```diff
@@ -0,0 +1,303 @@
...
+          maxPendingRequests: 8
+          maxRetries: 2
+          maxRequests: 2
+```
+
+We will apply the configuration with `kumactl apply -f [..]` or via the [HTTP API](/docs/{{ page.version }}/reference/http-api).
+{% endtab %}
+{% endtabs %}
+
+#### Outlier detection for inbound traffic to backend service
```
slonka, line 109: I know it's not strictly related to docs but how does outlier detection work on an inbound?

slonka (commented):
A couple of things:
1. How does it work for gRPC? 
2. Does the client need to specify the protocol (e.g. HTTP) for certain things to work?
3. I feel like you're listing all of the possible configuration options and I think we've agreed that this will be autogenerated from code and in docs we should focus on the configuration options that are harder to grasp and go into more details than in code.

I'll do another round of review probably tomorrow.

app/_src/policies/meshcircuitbreaker.md
bartsmykla: sgtm

app/_src/policies/meshcircuitbreaker.md
bartsmykla: I would say it's explained deep enough as the next ~10 lines are about outlier detection

app/_src/policies/meshcircuitbreaker.md
bartsmykla: addressed

app/_src/policies/meshcircuitbreaker.md
bartsmykla: addressed

app/_src/policies/meshcircuitbreaker.md
bartsmykla: addressed

app/_src/policies/meshcircuitbreaker.md
bartsmykla: local failures:
> Various locally originated problems are detected: timeout, TCP reset, ICMP errors, etc. This detection type is supported by http router and tcp proxy.

external errors:
> server-side error responses (for HTTP traffic, 5xx responses; for TCP traffic, connection failures; for Redis, failure to respond PONG; etc.) before a consecutive total failure ejection occurs.

app/_src/policies/meshcircuitbreaker.md
bartsmykla: I'm not sure what you mean by the first sentence, but for the second:
> Each worker thread maintains its own connection pools for each cluster, so if an Envoy has two threads and a cluster with both HTTP/1 and HTTP/2 support, there will be at least 4 connection pools.
> https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/connection_pooling#number-of-connection-pools

bartsmykla:
> 1. How does it work for gRPC?

There is no differentiation between gRPC and HTTP in this context, so it will behave the same as for HTTP2

> 2. Does the client need to specify the protocol (e.g. HTTP) for certain things to work?

What you mean by certain things?

> 3. I feel like you're listing all of the possible configuration options and I think we've agreed that this will be autogenerated from code and in docs we should focus on the configuration options that are harder to grasp and go into more details than in code.

To be honest I'm not sure how should I differentiate it here. Do you have some suggestions?

app/_src/policies/meshcircuitbreaker.md
bartsmykla: I think it's addressed, please reopen conversation if you think otherwise

slonka:
> There is no differentiation between gRPC and HTTP in this context, so it will behave the same as for HTTP2

If it works the same way for `gRPC` should we mention that?

> What you mean by certain things?

E.g. 

```
            totalFailures:
              consecutive: 10
```

and in docs we have: 

```
 The number of consecutive server-side error responses (for HTTP traffic, 5xx responses; for TCP
  traffic, connection failures ...
```

So I'm guessing is that if someone does not have `appProtocol` set then Kuma will only take into account connection failures and if they have `appProtocol=http` then it will count both connection failures and 5xx responses. Am I right? So "certain things" (configuration options) work (or behave differently) based on the protocol defined. In other policies we link to https://kuma.io/docs/2.0.x/policies/protocol-support-in-kuma/#protocol-support-in-kuma and we had reports from users not realising that they need this protocol to be set for some things to work (I remember I posted [this comment](https://github.com/kumahq/kuma/issues/1256#issuecomment-1199077656) because of that).

> To be honest I'm not sure how should I differentiate it here. Do you have some suggestions?

I agree it's hard to figure out. Until we have this autogenerated field docs I think we can leave it as it is. But for the future we should come up with some guideline which things are worth extra docs and which can be just an entry in the autogenerated list of fields. I feel like if something is just a variable then we can skip it, if there is an algorithm behind a field then this is a good candidate to get into more details. Like [rate limited backoff](https://kuma.io/docs/dev/policies/meshretry/#rate-limited-backoff) has a 3 step algorithm to decide if / when a retry is issued, and `numRetries: 10` is just a number so we don't have to go into too much details there.

app/_src/policies/meshcircuitbreaker.md
```diff
@@ -0,0 +1,305 @@
...
+| MeshSubset        | ✅         | ❌   | ❌    |
+| MeshService       | ✅         | ✅   | ❌    |
+| MeshServiceSubset | ✅         | ❌   | ❌    |
+| MeshGatewayRoute  | ❌         | ❌   | ❌    |
+
+To learn more about the information in this table, see the [matching docs](/docs/{{ page.version }}/policies/targetref).
+
+## Configuration
+
+### Examples
```
slonka, line 53: Usually all of the navbar entries are level 2 - https://github.com/kumahq/kuma-website/blame/18a1a3f4a97f482ef403231c5c7f42bc145c0b60/app/_src/policies/meshtrafficpermission.md#L31 and configuration has configuration options (the thing that starts [here](https://github.com/kumahq/kuma-website/pull/1214/files#diff-bf631074d8bc5d74e7ecab393d60df436a9379befbee4010871f84b639e5f0a1R195)). I don't see any reason to link in navbar to "configuration" if it's empty.

bartsmykla: that was just temporary, I included configuration section with the latest commit

app/_data/docs_nav_kuma_dev.yml
```diff
@@ -672,6 +672,17 @@ items:
             url: "/policies/meshproxypatch/#examples"
           - text: Merging
             url: "/policies/meshproxypatch/#merging"
+      - text: MeshCircuitBreaker (Beta)
+        url: /policies/meshcircuitbreaker/
+        items:
+          - text: TargetRef support matrix
+            url: "/policies/meshcircuitbreaker/#targetref-support-matrix"
+          - text: Examples
+            url: "/policies/meshcircuitbreaker/#examples"
```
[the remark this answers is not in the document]
bartsmykla: addressed

app/_src/policies/meshcircuitbreaker.md
```diff
@@ -0,0 +1,581 @@
...
+This policy provides 5 different types of detectors, and they are triggered on some deviations in the upstream service
+behavior. All detectors could coexist on the same outbound interface.
+
+Once one of the detectors has been triggered the corresponding data plane proxy is ejected from the set of the load
+balancer for a period equal to [baseEjectionTime](#outlier-detection). Every further ejection of the same data plane
+proxy will further extend the [baseEjectionTime](#outlier-detection) multiplied by the number of ejections: for example
+the 4th ejection will be lasting for a period of time of `4 * baseEjectionTime`.
+
+This policy provides **passive** checks.
+If you want to configure **active** checks, please utilize the [MeshHealthCheck](/docs/{{ page.version }}/policies/meshhealthcheckcz)
```
slonka, line 34: This is tripping the link checker

slonka (approved):
I think it's alright now.

app/_src/policies/meshcircuitbreaker.md
slonka: Well it would act more like a inbound rate limiter, but since it passed MADR then I'm not going to protest.

app/_src/policies/meshcircuitbreaker.md
bartsmykla: agot it, thanks for catching this!

PR #1218, opened 2023-01-27
chore(docs) cut off 2.1.x docs

Automaat:
Signed-off-by: Marcin Skalski <marcin.skalski@konghq.com>


app/_data/docs_nav_kuma_2.1.x.yml
```diff
@@ -0,0 +1,897 @@
+product: kuma
+release: 2.1.x
+items:
```
cloudjumpercat, line 3:
```suggestion
generate: true
items:
```
@Automaat I think we're missing this which is why it won't generate the nav

bartsmykla (approved):

PR #1230, opened 2023-02-03
fix: use 302 redirects for latest and pagination doc

lahabana:
Using 301 was causing browsers to cache the redirect which was causing problems when the latest version changed

### DIFF
--- a/app/_common_redirects
+++ b/app/_common_redirects
@@ -14,13 +14,13 @@
 /docs/:version/policies/virtual-outbounds/ /docs/:version/policies/virtual-outbound/ 301
 
 # Docs
-/docs/latest/documentation/gateway /docs/LATEST_RELEASE/explore/gateway 301
-/docs/latest/deployments /docs/:version/LATEST_RELEASE/deployments 301
-/docs/latest/documentation/deployments /docs/LATEST_RELEASE/introduction/deployments 301
-/docs/latest/explore/backends /docs/LATEST_RELEASE/documentation/configuration 301
-/docs/latest/security/zone-ingress-auth /docs/LATEST_RELEASE/security/zoneproxy-auth 301
-/docs/latest/security/zoneegress-auth /docs/LATEST_RELEASE/security/zoneproxy-auth 301
-/docs/latest/* /docs/LATEST_RELEASE/:splat 301
+/docs/:version/documentation/gateway /docs/:version/explore/gateway 301
+/docs/:version/deployments /docs/:version/:version/deployments 301
+/docs/:version/documentation/deployments /docs/:version/introduction/deployments 301
+/docs/:version/explore/backends /docs/:version/documentation/configuration 301
+/docs/:version/security/zone-ingress-auth /docs/:version/security/zoneproxy-auth 301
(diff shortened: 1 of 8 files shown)


app/_common_redirects
```diff
@@ -14,13 +14,13 @@
...
 # Docs
-/docs/latest/documentation/gateway /docs/LATEST_RELEASE/explore/gateway 301
-/docs/latest/deployments /docs/:version/LATEST_RELEASE/deployments 301
-/docs/latest/documentation/deployments /docs/LATEST_RELEASE/introduction/deployments 301
-/docs/latest/explore/backends /docs/LATEST_RELEASE/documentation/configuration 301
-/docs/latest/security/zone-ingress-auth /docs/LATEST_RELEASE/security/zoneproxy-auth 301
-/docs/latest/security/zoneegress-auth /docs/LATEST_RELEASE/security/zoneproxy-auth 301
-/docs/latest/* /docs/LATEST_RELEASE/:splat 301
+/docs/:version/documentation/gateway /docs/:version/explore/gateway 301
+/docs/:version/deployments /docs/:version/:version/deployments 301
```
fabianrbz, line 18: We need to keep `LATEST_RELEASE` here, it gets replaced by the [generator](https://github.com/kumahq/kuma-website/blob/master/app/_plugins/generators/redirects.rb#L57) with the latest version

fabianrbz (changes requested):

app/_common_redirects
lahabana: But the redirect applies to all versions no?

fabianrbz: The ones you changed applied only to the ones containing `/latest/`  in the segment and we should keep them, which I believe should have a `302` status code now.
Also, you are right, we should add the redirects for all versions like you did

lahabana: But I'd like `/docs/1.8.x/documentation/gateway` to be redirected to `/docs/1.8.x/explore/gateway` for example even though latest is 2.1.x.

I think latest would still work but with 2 redirects:

1. redirect `/docs/latest/documentation/gateway` --> 301 `/docs/latest/explore/gateway` This is defined line 17
2. redirect `/docs/latest/explore/gateway` --> 302 `/docs/LATEST_RELEASE/explore/gateway` This is defined line 23

Am I missing/misunderstanding something here?

fabianrbz: Sorry if wasn't clear, we need the following:

fabianrbz: Sorry if I wasn't clear. We need both the redirects for all the versions and the one from `latest` to `<latest release>`.
You are right your changes take care of that because of line [23](https://github.com/kumahq/kuma-website/pull/1230/files#diff-1d8d42485ba225179e544f4691efa102099ec8479b432239ec2cec494dde3082R23)  ( sorry I missed that earlier)

fabianrbz:
@lahabana should the changes made to the `size` be included in this PR?

fabianrbz (approved):

lahabana:
> @lahabana should the changes made to the size be included in this PR?
100% I'm just lazy sometimes... I'll do another PR

slonka (approved):

PR #1245, opened 2023-02-24
Create the Kuma in Production section

cloudjumpercat:
- Created the new Kuma in Production section in the `dev` nav based on the Kuma IA rework here: https://docs.google.com/spreadsheets/d/1bdYIFK4RdoOGfunEd1txj8se9bR_6F55aoP3r52rDjc/edit#gid=1809471940
- Moved and renamed files per the spreadsheet.
- Since they are single sourced, updated the 2.0 and 2.1 nav files to point to the files that I moved with src.
- Docs ticket: DOCU-2857

Did you sign your commit? Yes [Instructions](https://github.com/kumahq/.github/blob/main/CONTRIBUTING.md#sign-your-commits)


cloudjumpercat:
Marking as a draft because I still have some questions that need to be resolved before this can be merged.

app/_data/docs_nav_kuma_2.0.x.yml
```diff
@@ -90,28 +91,39 @@ items:
             url: "/quickstart/universal/#next-steps"
       - text: Standalone deployment
         url: /deployments/stand-alone/
+        src: /production/deployment/stand-alone/
         items:
```
cloudjumpercat, line 95: @fabianrbz In the 2.1 nav, I had to remove all the links to headers from pages that I moved because they resulted in errors when I ran make build. Ex. I had to remove "Failure modes" (src: /production/stand-alone/#failure-modes) for the Standalone deployment page. But I left them in the 2.0 nav as is and they were fine and displayed correctly?

app/_data/docs_nav_kuma_dev.yml
```diff
@@ -19,13 +19,95 @@ items:
...
+            url: /production/deployment/networking/
+      - title: Control plane deployment
+        group: true
+        items:
+          # - text: Kong Mesh license
+            # url: /production/
+          - text: Deploy a multi-zone global control plane
+            url: /production/cp-deployment/multi-zone/ 
+            group: true
+            items:
```
cloudjumpercat, line 49: @fabianrbz Do you know why the subnav items below "Deploy a multi-zone control plane" won't show on the site?

app/_data/docs_nav_kuma_dev.yml
```diff
@@ -19,13 +19,95 @@ items:
...
+              # - text: Add a zone
+                # url: /production/
+              # - text: Configure multi-zone authentication
+                # url: /production/
```
cloudjumpercat, line 53: @lahabana @johnharris85 I didn't have a link for these in the IA spreadsheet so I wasn't sure if we'd be creating these pages from scratch or if there were existing pages that cover these two topics.

app/_data/docs_nav_kuma_dev.yml
```diff
@@ -19,13 +19,95 @@ items:
...
+          - text: Multi-zone deployment
+            url: /production/deployment/multi-zone/
+          - text: Default ports
+            url: /production/deployment/networking/
+      - title: Control plane deployment
+        group: true
+        items:
+          # - text: Kong Mesh license
+            # url: /production/
+          - text: Deploy a multi-zone global control plane
```
cloudjumpercat, line 46: @fabianrbz Some of the items in the nav that are down a level don't have a title that displays for the page (I think this is one of them) even though the file has a title in the meta data. Do you know how I can fix that?

app/_data/docs_nav_kuma_dev.yml
```diff
@@ -19,13 +19,95 @@ items:
...
+                url: /production/cp-deployment/zone-ingress/
+              - text: Zone Egress
+                url: /production/cp-deployment/zoneegress/
+              - text: Configure zone proxy authentication
+                url: /production/cp-deployment/zoneproxy-auth/
+          - text: Deploy a standalone control plane
+            url: /production/stand-alone/
+          - text: Control plane configuration reference
+            url: /production/cp-config-reference/
+            src: /generated/kuma-cp/
```
cloudjumpercat, line 64: @fabianrbz I don't think I formatted the src of this correctly because it won't display.

app/_data/docs_nav_kuma_2.0.x.yml
```diff
@@ -19,6 +19,7 @@ items:
             url: "/introduction/how-kuma-works/#kuma-vs-xyz"
       - text: Deployments
         url: /introduction/deployments/
+        src: /production/deployment/
```
lahabana, line 22: I thought single source was only used from 2.1.x no? If so it seems like these page moves shouldn't happen in 2.0.x no?

app/_data/docs_nav_kuma_dev.yml
lahabana: I think we'd do it from scratch yes

app/_src/production/dp-config/cni.md
```diff
@@ -1,5 +1,6 @@
 ---
-title: Kuma CNI
+title: Configure the Kuma CNI
```
lahabana, line 2: Should we get rid of product names in titles? It makes it easier to reuse the sidebar between products (as we can't use `site.product_name` in the sidebar?

lahabana (commented):
Great to see this starting!

app/_data/docs_nav_kuma_2.0.x.yml
fabianrbz: @cloudjumpercat did you leave the hash portion of the URL in the `src`?
e.g. `/production/deployment/stand-alone/#standalone-deployment` should be `/production/deployment/stand-alone/`, it should point to the file name not the url

app/_data/docs_nav_kuma_dev.yml
fabianrbz: I think the problem with this one is that you set `generate: false` so it won't generate the page for that item, that should be ok but the thing is that the URL doesn't match the file so it will return a 404.

However, if you remove that, it fails because it will try to find `/generated/kuma-cp` under `app/_src` which isn't there.

I guess we have two options here: make the URL match the file path or add support for using files outside of `app/_src` in this case `app/docs/` as a fallback in case it doesn't find them.

app/_data/docs_nav_kuma_dev.yml
fabianrbz: I don't think we support so many nesting levels, we could add that though
You could add `title: Deploy a multi-zone global control plane` and they should show up but the navbar won't work properly

app/_data/docs_nav_kuma_dev.yml
fabianrbz: Same reason as above, we don't support so many nesting levels :(

app/_data/docs_nav_kuma_2.0.x.yml
cloudjumpercat: @fabianrbz I did. Ah, so if I do:
```
- text: About
            url: "/deployments/stand-alone/#standalone-deployment"
            src: /production/deployment/stand-alone/
```
That would take me to the correct section on that page?

fabianrbz: yes, `src` just tells the code to render that file for that page, the `hash` portion of the `url` will make the page scroll to the corresponding section

app/_data/docs_nav_kuma_dev.yml
cloudjumpercat: > make the URL match the file path

I was going to pick this option, but it looks like this is a generated file and I don't think it's listed in the nav in the current docs. The URL for this in the public docs is `/generated/kuma-cp/`, but that doesn't seem to work and I don't know what the actual file path is.

app/_src/production/dp-config/cni.md
cloudjumpercat: @lahabana I think that we have automation on the Mesh docs side that automatically replaces "kuma" in titles/nav/URLs with "kong mesh". @fabianrbz Is that accurate?

app/_data/docs_nav_kuma_dev.yml
fabianrbz: something like this should work:
```
- text: Control plane configuration reference
   url: /generated/kuma-cp/
   generate: false
```

app/_src/production/dp-config/cni.md
fabianrbz: I believe so

lahabana: Even in the nav? Last time I checked that wasn't the case for the nav. This was making things harder to potentially add scripting to sync automatically the sidebar instead of duplicating it like we do nowadays.

lahabana: I started working on an extension for the site-generator to handle inheriting a sidebar: https://github.com/Kong/jekyll-generator-single-source/pull/7

Unfortunately I'm a bit busy atm and don't have time to complete this work.

app/_src/production/deployment/multi-zone.md
```diff
@@ -0,0 +1,96 @@
...
+  - Receive XDS configuration from the local zone control plane.
+  - Proxy traffic from local data plane proxies:
+    - to zone ingress proxies from other zones;
+    - to external services from local zone;
+
+{% if_version gte:2.2.x %}
+
+## More information
+
+For more information about how to configure a multi-zone deployment, see [Deploy a multi-zone global control plane](/docs/{{ page.version }}/production/cp-deployment/multi-zone/).
```
cloudjumpercat, line 94: @fabianrbz Do you know why this link is failing the link checker? It looks like it's only failing for 2.1, so I made this only appear for 2.2 (because that's when I made this doc change), but it's still failing.

app/_data/docs_nav_kuma_2.1.x.yml
```diff
@@ -95,28 +96,39 @@ items:
...
       - text: Multi-zone deployment
         url: /deployments/multi-zone/
+        src: /production/deployment/multi-zone/
         items:
           - text: About
             url: "/deployments/multi-zone/#multi-zone-deployment"
+            src: /production/deployment/multi-zone/
           - text: Usage
             url: "/deployments/multi-zone/#usage"
+            src: /production/cp-deployment/multi-zone/
```
cloudjumpercat, line 122: @fabianrbz I think this one is causing the Netlify build to fail and I'm not sure why since it shouldn't be using the URL as a filename. Here's the error from the Netlify logs:

```
Error message
11:53:10 AM:   Deploy did not succeed: Invalid filename 'docs/2.1.x/deployments/multi-zone/#usage/index.html'. Deployed filenames cannot contain # or ? characters
```

fabianrbz: Looks like this is a weird edge case for the generator.
The problem is how the generator interprets the different items. It works like this (simplified version):
* if the item has a `src` it will generate the page based on that file and use the `url` as the file name of the page
* if it doesn't have a `src` but it does have a `url`, it uses the `url` as the source and generates the page unless the URL has a `hash` fragment. In that case, it assumes that the link is just a reference to a section of the page, so there's no need to generate anything, it should be generated by a different item.
Like in [this](https://github.com/Kong/docs.konghq.com/blob/main/app/_data/docs_nav_gateway_3.2.x.yml#L564-L585) case, the first item generates the page, the others skip the page generation and point to sections of the page.

One problem is that you are using a `hash` fragment AND a `src` at the same time, so the generator will generate a page for that item, and the `hash` ends up being a part of the file name of the page. The other problem is that you are using two different files as the source for the same URL, it should be one file for each URL.

Fortunately, there's a simple fix.
[Here](https://github.com/kumahq/kuma-website/blob/7cd0e2deb25df964fdebdbcf65f7699987440f25/app/_data/docs_nav_kuma_2.1.x.yml#L113-L131) the top level item can generate the page based on `src: /production/cp-deployment/multi-zone/` and the subitems can drop the `src`. This tells the generator to generate the page based on that `src` for the top-level item, and skip the generation for the subitems. They will be just links with a `hash` fragment that link to sections of the page.

cloudjumpercat:
❗ This shouldn't be merged until the corresponding Mesh PR (I'll link it here once it's created) is merged otherwise it will break docs in the Mesh side.

slonka:
@lahabana can you re-do the review?

lahabana (approved):

slonka (dismissed):

app/_data/docs_nav_kuma_dev.yml
```diff
@@ -129,63 +174,18 @@ items:
...
-        url: /explore/dpp/
-        items:
-          - text: Concepts
-            url: "/explore/dpp/#concepts"
-          - text: Dataplane entity
-            url: "/explore/dpp/#dataplane-entity"
-          - text: Dynamic configuration of the data plane proxy
-            url: "/explore/dpp/#dynamic-configuration-of-the-data-plane-proxy"
-      - text: Data plane on Kubernetes
-        url: /explore/dpp-on-kubernetes/
```
slonka, line 142: these no longer work - https://github.com/kumahq/kuma-website/actions/runs/4362925933/jobs/7628399011

slonka (commented):
Please fix the links from the link checker - https://github.com/kumahq/kuma-website/actions/runs/4362925933/jobs/7628399011

app/_redirects
```diff
@@ -34,3 +34,33 @@ https://prometheus.metrics.kuma.io/port /docs/LATEST_RELEASE/reference/kubernete
...
 https://traffic.kuma.io/exclude-inbound-ports /docs/LATEST_RELEASE/reference/kubernetes-annotations/#traffic-kuma-io-exclude-inbound-ports/ 302
 https://traffic.kuma.io/exclude-outbound-ports /docs/LATEST_RELEASE/reference/kubernetes-annotations/#traffic-kuma-io-exclude-outbound-ports/ 302
+
+# kuma IA redirects
+/docs/:version/introduction/deployments/  /docs/:version/production/deployment/
+/docs/:version/deployments/stand-alone/ /docs/:version/production/deployment/stand-alone/
+/docs/:version/deployments/multi-zone/  /docs/:version/production/deployment/multi-zone/
+/docs/:version/deployments/multi-zone/#usage  /docs/:version/production/cp-deployment/multi-zone/
+/docs/2.1.x/deployments/multi-zone/#usage /docs/2.1.x/production/cp-deployment/multi-zone/
+/docs/2.1.x/deployments/stand-alone/#usage  /docs/2.1.x/production/cp-deployment/stand-alone/
```
cloudjumpercat, line 44: @fabianrbz These ones still aren't working, but because I split the page, I think the "2.1.x" has to be in the URL since 2.1.x should be the only impacted version.

fabianrbz:
@cloudjumpercat kuma does support `/latest/` but we redirect them to the corresponding version [here](https://github.com/kumahq/kuma-website/blob/master/app/_redirects#L10).

From the logs: 
* Redirecting `/docs/latest/policies/mesh/` to `/docs/latest/production/mesh/` added [here](https://github.com/kumahq/kuma-website/pull/1245/files#diff-a19a552735ad7e7a99a125b16220722ae6a771a68b5b06332243b6709a357b4bR50)
* Redirecting `/docs/latest/production/mesh/` to `/docs/2.1.x/production/mesh/` expected based on the `latest` -> `version`

Which means that the redirects are working.

It looks like the file `app/_src/policies/mesh.md` was renamed to `app/_src/production/mesh.md`. However, the URL hasn't changed, at least [not for 2.1.x](https://github.com/kumahq/kuma-website/blob/66a0ec1b33075a5c1538fba28bb3fba4ab5263d1/app/_data/docs_nav_kuma_2.1.x.yml#L447-L448) and the previous versions.  So all those work just fine, they don't need/use the redirect.
However, It did change for `dev` though, which means that for the next release, it will make sense to have the redirect (we should also add the correct canonical URLs but that is a different story). 
We could remove it and create a ticket or something so that we remember to add it back for the next release.

It fails now because `latest` is `2.1.x` and that version still generates the old URL, but because of the redirects chain mentioned above, it ends up generating the wrong URL and returning a 404.
Does that make sense?

fabianrbz:
@cloudjumpercat that's correct

app/_redirects
```diff
@@ -38,8 +38,6 @@ https://traffic.kuma.io/exclude-inbound-ports /docs/LATEST_RELEASE/reference/kub
 https://traffic.kuma.io/exclude-outbound-ports /docs/LATEST_RELEASE/reference/kubernetes-annotations/#traffic-kuma-io-exclude-outbound-ports/ 302
```
cloudjumpercat, line 40:
```suggestion
# kuma IA redirects
/docs/:version/deployments/multi-zone/#usage /docs/:version/production/cp-deployment/multi-zone/
/docs/:version/deployments/stand-alone/#usage  /docs/:version/production/cp-deployment/stand-alone/
```

cloudjumpercat:
@slonka Could you review this again when you have the time? The only failing link is the one you mentioned ( https://chat.kuma.io/) that would be expected to fail.

slonka (approved):

PR #1271, opened 2023-04-03
sync versions.yml from kumahq/kuma

lahabana:


lahabana:
Docs are already removed in KM

app/_data/versions.yml
```diff
@@ -1,34 +1,61 @@
...
+  release: 1.4.x
+  releaseDate: "2021-11-22"
+  endOfLifeDate: "2022-11-22"
+  branch: release-1.4
+- edition: kuma
+  version: 1.5.4
+  release: 1.5.x
+  releaseDate: "2022-02-24"
+  endOfLifeDate: "2023-02-24"
+  branch: release-1.5
```
slonka, line 24: should we be including EOL versions?

lahabana: Yes because we still keep docs around for EOL versions.

lahabana: Not real sure what's the reason for this but it's not something I want to fix now :)

PR #1276, opened 2023-04-03
Meshgateway new docs

johnharris85:
Docs for https://github.com/kumahq/kuma/pull/6388


slonka (commented):
In general looks good but I think you need to gate the new stuff with 

```
{% if_version gte:2.2.x %}
```

otherwise it will be displayed on older versions of the docs.

slonka (approved):

PR #1289, opened 2023-04-06
feat: add policy schema display

michaelbeaumont:
[Rendered](https://deploy-preview-1289--kuma.netlify.app/docs/2.1.x/policies/meshcircuitbreaker/#complete-spec)

Closes #1274

PR #1290, opened 2023-04-06
docs(policy): automatically convert policy yaml

jakubdyszkiewicz:
Add plugin to generate policy tabs for universal and kubernetes and automatically convert universal yaml to kubernetes yaml.

I changed only one file, but once this is merged I'll go over all occurrences in the project to clean it up.

The difference with MeshCircuitBreaker is that we are missing
```
We will apply the configuration with `kubectl apply -f [..]`.
```
and
```
We will apply the configuration with `kumactl apply -f [..]` or via the [HTTP API](/docs/{{ page.version }}/reference/http-api).
```

I don't find this useful, but I can also add this information to an HTML generated by the plugin. It can be even opt-in.


lahabana:
@fabianrbz for this to be usable in both Kuma and Kong-Mesh we need to extract this in a gem?

bartsmykla (approved):

Automaat:
I am missing one thing here. I think it is beneficial to have policies that have service names easily understandable on every platform. For example on universal you can set service name to any value like: `demo-app`, on k8s on the other hand service name contains namespace and port: `demo-app_kuma-demo_svc_5000`. It would be good to include it in this plugin somehow.

Another topic I think it would be also helpful for users to base all policies example on demo app, so they can simply play with them and apply them without changing anything.

michaelbeaumont:
@jakubdyszkiewicz is this already working with the new reusable plugin stuff in https://github.com/kumahq/kuma-website/pull/1303?

jakubdyszkiewicz:
@michaelbeaumont It doesn't. Is there a repo or something to this plugin to the gem?

michaelbeaumont:
@jakubdyszkiewicz It's in this repo, I think you just have to make changes similar to https://github.com/kumahq/kuma-website/pull/1289/commits/46e2f015d4aa252fa8c9c6c874ffb62df20466ad

jakubdyszkiewicz:
It seems it's there, Charly moved this around

michaelbeaumont (approved):
Nice! LGTM

PR #1300, opened 2023-04-12
fix(installer.sh): make preview work correctly

lahabana:
Fix https://github.com/Kong/kong-mesh/issues/3152

### DIFF
--- a/app/installer.sh
+++ b/app/installer.sh
@@ -104,12 +104,6 @@ if [ -z "$VERSION" ]; then
   fi
 fi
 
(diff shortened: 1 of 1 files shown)


slonka (approved):
Seems to work (check by doing `curl -L https://deploy-preview-1300--kuma.netlify.app/installer.sh | VERSION=0.0.0-preview.v853fd7b40 sh -`). How come this was not caught by CI? Are we not testing `previews` in CI? If so please add a test or create an issue to add a test.

PR #1312, opened 2023-04-19
fix(_redirects): correct redirects following IA work

lahabana:
IA was missing a bunch of redirects to avoid broken links.


lahabana:
FYI @cloudjumpercat

PR #1314, opened 2023-04-19
fix: kumactl install docs should be more generic

lahabana:
This makes things simpler in inherited docs


bartsmykla (approved):
I don't know ruby, but it looks good to me from the logical standpoint

PR #1365, opened 2023-06-21
feat(annotations): add init-first

johnharris85:
Docs for https://github.com/kumahq/kuma/pull/7070

### DIFF
--- a/app/_src/reference/kubernetes-annotations.md
+++ b/app/_src/reference/kubernetes-annotations.md
@@ -685,7 +685,7 @@ spec: ...
 ### `kuma.io/init-first`
 
 Allows specifying that the {{site.mesh_product_name}} init container should run first (ahead of any other init containers). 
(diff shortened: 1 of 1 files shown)


app/_src/reference/kubernetes-annotations.md
```diff
@@ -681,3 +681,21 @@ metadata:
     kuma.io/sidecar-drain-time: "10s"
 spec: ...
 ```
+{% if_version gte:2.3.x %}
+### `kuma.io/init-first`
+
+Allows specifying that the {{site.mesh_product_name}} init container should run first (ahead of any other init containers). 
+The default is `false` if omitted. Setting this to true may be desirable for security, as it would prevent network access for other init containers. The order is _not_ guaranteed, as other mutating admission webhooks may further manipulate this ordering.
```
bartsmykla, line 688:
```suggestion
The default is `false` if omitted. Setting this to `true` may be desirable for security, as it would prevent network access for other init containers. The order is _not_ guaranteed, as other mutating admission webhooks may further manipulate this ordering. 
```

bartsmykla (approved):
One nit, but LGTM overall

PR #1377, opened 2023-06-23
docs: improve systemd deployment docs

bartsmykla:
* Instead of hardcoding `kuma` in systemd code blocks now we are using `{{ site.mesh_product_name | downcase }}`.
* Added color scheme for `systemd` code blocks for `Prism` library
* Made spacing consistent in interpolations (there is now a space after `{{` and before `}}` (i.e. `{{ site.mesh_product_name }}` instead of `{{site.mesh_product_name}}`).
* Made spacing in `systemd` unit code blocks consistent (i.e. there is now always `Restart = always` instead of `Restart=always` in some places and `Restart = always` in other
* Changed `pgrep kuma-cp` to `$(pgrep kuma-cp)` as command the former format of command substitution is deprecated for the long time already.
* Capitalized first letters in the Description block of `kuma-dp` systemd unit, to be consistent with the `kuma-cp` block.
* Used product name in the code block instead of hardcoded `kuma`.
* Make path to binaries consistent in `kuma-cp` and `kuma-dp` systemd unit code blocks. This path needs to be updated by the user, so having imaginary path like `/home/kuma/kuma-2.2.2/bin/` is fine I think.

### DIFF
--- a/app/_src/production/cp-deployment/systemd.md
+++ b/app/_src/production/cp-deployment/systemd.md
@@ -15,9 +15,9 @@ After = network.target
 Documentation={{ site.links.web }}
 
 [Service]
-User = {{ site.mesh_product_name_lowercase }}
+User = {{ site.mesh_product_name | downcase }}
 Environment = KUMA_MODE=standalone
-ExecStart = /path/to/{{ site.mesh_product_name_lowercase }}/bin/kuma-cp run --config-file=/home/{{ site.mesh_product_name_lowercase }}/cp-config.yaml
+ExecStart = /path/to/{{ site.mesh_product_name | downcase }}/bin/kuma-cp run --config-file=/home/{{ site.mesh_product_name | downcase }}/cp-config.yaml
 # if you need your Control Plane to be able to handle a non-trivial number of concurrent connections
 # (a total of both incoming and outgoing connections), you need to set proper resource limits on
 # the `kuma-cp` process, especially maximum number of open files.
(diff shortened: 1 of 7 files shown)


app/_src/production/cp-deployment/systemd.md
```diff
@@ -3,21 +3,21 @@ title: Systemd
...
+Description = {{ site.mesh_product_name }} Control Plane
 After = network.target
-Documentation=https://kuma.io
+Documentation={{ site.links.web }}
 
 [Service]
-User=kuma
-Environment=KUMA_MODE=standalone
-ExecStart = /path/to/kuma/bin/kuma-cp run --config-file=/home/kuma/cp-config.yaml
+User = {{ site.mesh_product_name_lowercase }}
```
slonka, line 18: as discussed on slack should we just use `{{ site.mesh_product_name | downcase }}`?

app/_src/production/cp-deployment/systemd.md
```diff
@@ -36,38 +36,38 @@ ExecStart = /path/to/kuma/bin/kuma-cp run --config-file=/home/kuma/cp-config.yam
...
-Documentation=https://kuma.io
+Description = {{ site.mesh_product_name }} Data Plane Proxy
+After = network.target
+Documentation = {{ site.links.web }}
 
 [Service]
-User=kuma
-ExecStart=/home/kuma/kong-mesh-1.9.1/bin/kuma-dp run \
+User = {{ site.mesh_product_name_lowercase }}
+ExecStart = /home/{{ site.mesh_product_name_lowercase }}/{{ site.mesh_product_name_lowercase }}-{{ page.latest_version }}/bin/kuma-dp run \
```
slonka, line 60: `page.latest_version` won't work for dev / current HEAD but I guess this is fine, people can replace that.

slonka (commented):
I'd probably go with `| downcase`, otherwise looks fine.

app/_src/production/cp-deployment/systemd.md
bartsmykla: I agree, this is just example, you still need to provide your dp template as well as the token.

app/_src/production/cp-deployment/systemd.md
bartsmykla: I actually like `| downcase` as it doesn't involve in making sure this variable is set in other places.

bartsmykla: Addressed

slonka (approved):

PR #1391, opened 2023-06-27
chore: vale vocabulary related improvements

bartsmykla:
Improved `vale` vocabulary dictionary:
- Deleted duplicated entries
- Combined multiple similar entries into single regular expression entry
- Added few missing words
- Sorted entries alphabetically
- Changed `authnz` to `auth[nz]` as there is no usage of `authnz` in the docs, but there are usages of `authn` and `authz` + I'm not sure if we should allow for usage of the `authnz`, which I believe is very specific, and I would prefer to see `authn/authz` than `authnz` as I believe it's easier to understand without focusing on particular word.

Did you sign your commit? [Instructions](https://github.com/kumahq/.github/blob/main/CONTRIBUTING.md#sign-your-commits)?: 👍 


.github/styles/Vocab/Base/accept.txt
```diff
@@ -142,39 +136,18 @@ uncheck
...
-url
-urls
+urls?
 uuid
 Valero
-validator
-validators
+validators?
+VMs?
 Vuejs
```
bartsmykla, line 146: I didn't want to change it without updating docs, but I believe we should allow `Vue.js` as it's the correct project name, and disallow `Vuejs`.

michaelbeaumont (approved):
Cool! Didn't know you can have regex tbh

PR #1394, opened 2023-06-28
docs(MeshAccessLog): add info about targetRef MeshGateway

michaelbeaumont:
[Can we display the info like this?](https://deploy-preview-1394--kuma.netlify.app/docs/dev/policies/meshaccesslog/#targetref-support-matrix)


lobkovilya (approved):

lahabana:
I think the idea is good. Would be nice down the road to have this table be a plugin as we're for sure growing in complexity here :)

PR #1442, opened 2023-08-22
fix(MeshGateway): fix typo on example tags table

nicoche:


michaelbeaumont (approved):
Thanks!

lahabana:
> I haven't figured out how to generate changes to ./app/docs/<VERSION> though 😕

We don't really bother updated the older versions of the doc now that we have multi-version based docs. Thanks @nicoche

PR #1469, opened 2023-09-22
docs(dns): mention DNS standards RFC in dns documentation

gperanich:
Adding reference to DNS standards as per ([RFC1035](https://www.ietf.org/rfc/rfc1035.txt)).

### DIFF
--- a/app/_src/networking/dns.md
+++ b/app/_src/networking/dns.md
@@ -119,8 +119,9 @@ Consuming a service handled by {{site.mesh_product_name}} DNS, whether from {{si
 <kuma-enabled-pod>$ curl http://echo-server_echo-example_svc_1010.mesh:80
 <kuma-enabled-pod>$ curl http://echo-server_echo-example_svc_1010.mesh
 ```
+Be aware of DNS standards ([RFC1035](https://www.ietf.org/rfc/rfc1035.txt)) if using underscores in the service name.
 
-A DNS standards compliant name is also available, where the underscores in the service name are replaced with dots. For example:
+A DNS standards-compliant name is also available, where the underscores in the service name are replaced with dots. For example:
 
 ```
 <kuma-enabled-pod>$ curl http://echo-server.echo-example.svc.1010.mesh:80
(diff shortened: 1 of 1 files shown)


app/_src/networking/dns.md
```diff
@@ -119,8 +119,9 @@ Consuming a service handled by {{site.mesh_product_name}} DNS, whether from {{si
 <kuma-enabled-pod>$ curl http://echo-server_echo-example_svc_1010.mesh:80
 <kuma-enabled-pod>$ curl http://echo-server_echo-example_svc_1010.mesh
 ```
+Be aware of DNS standards ([RFC1035](https://www.ietf.org/rfc/rfc1035.txt)) if using underscores in the service name.
 
-A DNS standards compliant name is also available, where the underscores in the service name are replaced with dots. For example:
+A DNS standards-compliant name is also available, where the underscores in the service name are replaced with dots. For example:
```
bartsmykla, line 124: * I removed the word "standards-compliant" because it is redundant. The word "DNS" already implies that the name complies with the DNS standards.
* I added the word "can" before "be used" to make it clear that this is just one option.
* I added the word "the" before "service name" for consistency.
```suggestion
A DNS-compliant ([RFC 1035](https://www.ietf.org/rfc/rfc1035.txt) name can also be used, where the underscores in the service name are replaced with dots. For example:
```

app/_src/networking/dns.md
```diff
@@ -119,8 +119,9 @@ Consuming a service handled by {{site.mesh_product_name}} DNS, whether from {{si
 <kuma-enabled-pod>$ curl http://echo-server_echo-example_svc_1010.mesh:80
 <kuma-enabled-pod>$ curl http://echo-server_echo-example_svc_1010.mesh
 ```
+Be aware of DNS standards ([RFC1035](https://www.ietf.org/rfc/rfc1035.txt)) if using underscores in the service name.
```
bartsmykla, line 122: I don't think this sentence brings any value, especially when we add the link to RFC to below section. If you are looking for DNS-compliant variant, there is assumption you are aware that your own service name has also follow the standard.
```suggestion
```

bartsmykla (commented):
Thank you for your contribution. I have few suggestions

bartsmykla (approved):

PR #1484, opened 2023-10-12
docs(certs): Add note about supported cert types to docs

cloudjumpercat:
The support team created [this KB](https://support.konghq.com/support/s/article/Kong-Mesh-Error-unable-to-parse-bytes-as-PEM-block-when-configuring-a-Control-Plane-certificates) and I thought that we should add this info to the docs. I didn't know what cert types are supported, but I would also like to have that added as well to help with troubleshooting.


app/_src/production/secure-deployment/certificates.md
```diff
@@ -157,25 +157,23 @@ The API Server HTTPS server is secured by default by autogenerated certificates.
...
-  --key-file=/tmp/tls.key
-```
-
-Since "tls.crt" is a self-signed cert, it is also a CA:
-```sh
-cp /tmp/tls.crt /tmp/ca.crt
-```
-
-2) Configure the control plane with generated certificates
+1. Prepare certificates. Generate TLS pair with a PKI of your choice and store it in PEM-encoded format in `/tmp/tls.crt`, `/tmp/tls.key`. Store the CA that was used to sign this pair in `/tmp/ca.crt`
```
cloudjumpercat, line 160: I made some formatting updates, but I'm not sure if these will render correctly since I couldn't check it out locally.

app/_src/production/secure-deployment/certificates.md
```diff
@@ -203,15 +201,21 @@ KUMA_API_SERVER_HTTPS_TLS_CERT_FILE=/tmp/tls.crt \
...
 {% endtabs %}
 
-3) Configure secure connection using `kumactl` CLI tool:
-```sh
-kumactl config control-planes add \
-  --name=<NAME> \
-  --address=https://<KUMA_CP_DNS_NAME>:5682 \
-  --ca-cert-file=/tmp/ca.crt \
-```
+  If you recieve an error like the following, make sure you are using a supported certificate type and that the certificate doesn't contain incomplete or corrupted data:
```
cloudjumpercat, line 204: Do you know what cert types are supported? I'd like to list them here.

app/_src/production/secure-deployment/certificates.md
cloudjumpercat: Also, I'm wondering if this note should also be added to the previous section (Encrypted communication) as well.

slonka: > I made some formatting updates, but I'm not sure if these will render correctly since I couldn't check it out locally.

you can always check the deployment preview https://deploy-preview-1484--kuma.netlify.app/docs/dev/production/secure-deployment/certificates/

app/_src/production/secure-deployment/certificates.md
slonka: From what I can see it's only PEM - https://github.com/kumahq/kuma/blob/fc86a9d3ae1fa14fe0a223dc74103a4948f8b83e/app/kuma-dp/pkg/dataplane/envoy/remote_bootstrap.go#L69

slonka (approved):

app/_src/production/secure-deployment/certificates.md
cloudjumpercat: Thanks! I fixed the formatting

app/_src/production/secure-deployment/certificates.md
```diff
@@ -201,21 +206,21 @@ KUMA_API_SERVER_HTTPS_TLS_CERT_FILE=/tmp/tls.crt \
...
 
-  If you recieve an error like the following, make sure you are using a supported certificate type and that the certificate doesn't contain incomplete or corrupted data:
+  If you receive an error like the following, make sure you are using a supported PEM certificate and that the certificate doesn't contain incomplete or corrupted data:
   ```sh
   Warning  FailedCreate  3m39s (x18 over 14m)  replicaset-controller  Error creating: Internal error occurred: failed calling webhook "namespace-kuma-injector.kuma.io": could not get REST client: unable to load root certificates: unable to parse bytes as PEM block
   ```
 
 
-3. Configure secure connection using `kumactl` CLI tool:
+3) Configure secure connection using the `kumactl` command-line tool:
```
slonka, line 215: I think you could use CLI but we need to define it here https://github.com/kumahq/kuma-website/blob/2359791dce1109e75a4dbf3fe1f1f65e98a5a78e/.github/styles/Vocab/Base/accept.txt#L19

slonka (approved):

PR #1490, opened 2023-10-20
fix: make port 22 in transparent-proxy clear

lahabana:
<Explain your change!>

### DIFF
--- a/app/_src/production/dp-config/transparent-proxying.md
+++ b/app/_src/production/dp-config/transparent-proxying.md
@@ -51,11 +51,23 @@ The host that will run the `kuma-dp` process in transparent proxying mode needs 
      --redirect-dns
    ```
 
+
 {% warning %}
 Please note that this command **will change** the host `iptables` rules.
+It's recommended to install the transparent proxy in a virtual network to easily recreate things from scratch.
 {% endwarning %}
 
-The changes will persist over restarts, so this command is needed only once. Reverting to the original state of the host can be done by issuing `kumactl uninstall transparent-proxy`.
+
+The changes will persist over restarts, so this command is needed only once. Reverting to the original state of the host can be done by issuing `kumactl uninstall transparent-proxy` {% if_version gte:2.2.x %} (which is not implemented yet see issue [#8071](https://github.com/kumahq/kuma/issues/8071)).{% endif_version %}
+
+{% warning %}
+If you are running transparent proxy on a host that is access through ssh you should exclude the ssh port to avoid being locked outside when applying iptables.
+You can do this with:
+
+```shell
+kumactl install transparent-proxy --exclude-inbound-ports 22 --exclude-outbound-ports 1521
+```
+{% warning %}
 
 ### Data plane proxy resource
 


app/_src/production/dp-config/transparent-proxying.md
```diff
@@ -51,11 +51,23 @@ The host that will run the `kuma-dp` process in transparent proxying mode needs
      --redirect-dns
    ```
 
+
```
slonka, line 54: nitpick:

```suggestion
```

app/_src/production/dp-config/transparent-proxying.md
```diff
@@ -51,11 +51,23 @@ The host that will run the `kuma-dp` process in transparent proxying mode needs
...
 {% warning %}
 Please note that this command **will change** the host `iptables` rules.
+It's recommended to install the transparent proxy in a virtual network to easily recreate things from scratch.
 {% endwarning %}
 
-The changes will persist over restarts, so this command is needed only once. Reverting to the original state of the host can be done by issuing `kumactl uninstall transparent-proxy`.
+
```
slonka, line 60: nitpick:

```suggestion
```

app/_src/production/dp-config/transparent-proxying.md
```diff
@@ -51,11 +51,23 @@ The host that will run the `kuma-dp` process in transparent proxying mode needs
...
+The changes will persist over restarts, so this command is needed only once. Reverting to the original state of the host can be done by issuing `kumactl uninstall transparent-proxy` {% if_version gte:2.2.x %} (which is not implemented yet see issue [#8071](https://github.com/kumahq/kuma/issues/8071)).{% endif_version %}
```
slonka, line 61: It only makes sense when you know that we rewrote TP. Why would we mention a command that does not exist.

I'd go with:

```suggestion
The changes will persist over restarts, so this command is needed only once. {% if_version lte:2.1.x %} Reverting to the original state of the host can be done by issuing `kumactl uninstall transparent-proxy`.{% endif_version %}{% if_version gte:2.2.x %} Reverting to the original state of the host is not yet implemented, see issue [#8071](https://github.com/kumahq/kuma/issues/8071).{% endif_version %}
```

slonka (commented):
I'd change one line mentioned.

app/_src/production/dp-config/transparent-proxying.md
bartsmykla: Wait.`kumactl install transparent-proxy` needs to be executed after every system reboot or you need to have some solution for iptables rules persistence in place. Otherwise, the transparent proxy configuration won't survive restarts.

bartsmykla: Wait. `kumactl install transparent-proxy` needs to be executed after every system reboot or you need to have some solution for iptables rules persistence in place. Otherwise, the transparent proxy configuration won't survive restarts.

bartsmykla (changes requested):

PR #1505, opened 2023-11-03
fix: sidebar being too tall

### LINKED ISSUE #1321: Scroll menu on docs sidebar too tall
### What happened?

Given the header size the .sidebar > .sidebar-links menu is larger than the visible area, 100vh assumes that it's the size of the whole screen when it's really 100vh - (size of headers) or approximately 85vh. At it's current 100vh when it overflows the screen you cannot scroll to the bottom.  

You can see this issue by going to https://kuma.io/docs/2.2.x/policies/meshloadbalancingstrategy/ and trying to scroll to the bottom of the sidebar menu.

I have tried 85vh and it appears to work however this should be tested with more screens. A more ideal path would be to use a calculation for this height.

kleinfreund:
## Changes

Fixes the sidebar being too tall which messes up the scrolling behavior: one currently often needs to scroll all the way to the bottom of the page to see the bottom items of the sidebar. This PR changes the sidebar behavior so that the sidebar can always be scrolled individually regardless of the main content height.

Resolves #1321.

## Notes

I made some development setup-related changes that I ran into while working on this.

One note on the development setup using `make run`: I had trouble getting things working and ultimately, with the help of @fabianrbz, got a working development server working by removing `netlify: sleep 50 && netlify dev` from the `Procfile`. I haven’t committed this change here because I’m not entirely sure what it implies.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ endif
 
 # Installs npm packages and gems.
 install: ruby-version-check
-	npm install -g netlify-cli
+	npm install -g netlify-cli@16.5.1
 	yarn install
 	bundle install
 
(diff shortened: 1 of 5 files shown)


michaelbeaumont (approved):
LGTM, seems to fix it

Getting to the bottom of the scrollbar on
https://kuma.io/docs/2.4.x/production/cp-deployment/stand-alone/
vs https://deploy-preview-1505--kuma.netlify.app/docs/2.4.x/production/cp-deployment/stand-alone/

Issue #1475, opened 2023-10-03
add docs for advanced localityAware LB


lahabana:
### Description

Once: https://github.com/kumahq/kuma/issues/2689 is implemented we should add docs.

PR #1506, opened 2023-11-03
feat(mlbs): documentation of advanced locality awareness

lukidzi:
Rendered: https://deploy-preview-1506--kuma.netlify.app/docs/dev/policies/meshloadbalancingstrategy/#advanced-localityawareness

xref: https://github.com/kumahq/kuma-website/issues/1475


app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -30,6 +30,38 @@ Locality-aware load balancing is enabled by default unlike its predecessor [loca
 - **`disabled`** – (optional) allows to disable locality-aware load balancing. When disabled requests are distributed 
 across all endpoints regardless of locality.
 
+{% if_version gte:2.5.x %}
+### Advanced LocalityAwareness
+Advanced locality-aware load balancing provides a more robust and straightforward method for balancing traffic within and across zones. This not only allows you to route traffic to a cross zone when the local zone service is unhealthy but also enables you to define traffic prioritization within the local zone and set cross-zone fallback priorities.
```
slonka, line 35: To me "cross zone" is not the right phrase to use. I'd go with:

```suggestion
Advanced locality-aware load balancing provides a more robust and straightforward method for balancing traffic within and across zones. This not only allows you to route traffic across zones when the local zone service is unhealthy but also enables you to define traffic prioritization within the local zone and set cross-zone fallback priorities.
```

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -30,6 +30,38 @@ Locality-aware load balancing is enabled by default unlike its predecessor [loca
...
+
+#### LocalZone
+Local zone routing allows you to define traffic routing rules within a local zone, prioritizing data planes based on tags and their associated weights. This enables you to allocate specific traffic percentages to data planes with particular tags within the local zone. If there are no healthy endpoints within the highest priority group, the next priority group takes precedence. Locality awareness within the local zone relies on tags within inbounds, so it's crucial to ensure that the tags used in the policy are defined for the service (Dataplane object on Universal, Deployment annotations on Kubernetes).
+
+- **`localZone`** - (optional) allows to define load balancing priorities between dataplanes in the local zone. When not defined traffic is distributed equally to all endpoints within the local zone.
```
slonka, line 40:
```suggestion
- **`localZone`** - (optional) allows to define load balancing priorities between dataplanes in the local zone. When not defined, traffic is distributed equally to all endpoints within the local zone.
```

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -30,6 +30,38 @@ Locality-aware load balancing is enabled by default unlike its predecessor [loca
...
+  - **`affinityTags`** - list of tags and their weights based on which traffic is loadbalanced
+    - **`key`** - defines tag for which affinity is configured. The tag needs to be configured on the inbound of the service. In case of Kubernetes pod needs to have a tag and it should be propagated. On Universal user needs to define it on the inbound of the service. In case there so match, the dataplane match to the last group with the lowest weight.
```
slonka, line 42: > should be propagated

what do you mean by that? propagated where, by what? propagated to all pods? if it's a strict requirement use `needs to be`, should implies that it's optional - https://datatracker.ietf.org/doc/html/rfc2119 

nitpick: comma missing between "case of Kubernetes, pod needs to"

> In case there so match

did you mean "in case there is no match"?

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -30,6 +30,38 @@ Locality-aware load balancing is enabled by default unlike its predecessor [loca
...
+
+#### LocalZone
+Local zone routing allows you to define traffic routing rules within a local zone, prioritizing data planes based on tags and their associated weights. This enables you to allocate specific traffic percentages to data planes with particular tags within the local zone. If there are no healthy endpoints within the highest priority group, the next priority group takes precedence. Locality awareness within the local zone relies on tags within inbounds, so it's crucial to ensure that the tags used in the policy are defined for the service (Dataplane object on Universal, Deployment annotations on Kubernetes).
+
+- **`localZone`** - (optional) allows to define load balancing priorities between dataplanes in the local zone. When not defined traffic is distributed equally to all endpoints within the local zone.
+  - **`affinityTags`** - list of tags and their weights based on which traffic is loadbalanced
```
slonka, line 41: shouldn't this be

```suggestion
  - **`affinityTags`** - list of tags and their weights based on which traffic is load balanced
```

? since we use "load balancing" (with a space)

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -30,6 +30,38 @@ Locality-aware load balancing is enabled by default unlike its predecessor [loca
...
+    - **`weight`** - (optional) weight of the tag used for load balancing. The bigger the weight the higher number of requests is routed to dataplanes with specific tag. By default we will adjust them so that 90% traffic goes to first tag, 9% to next, and 1% to third and so on.
+
+#### CrossZone
+{% warning %}
+Remember that cross-zone traffic requires [mTLS enabled](/docs/{{ page.version}}/policies/mutual-tls).
+{% endwarning %}
+Advanced locality-aware load balancing provides a powerful means of defining how your service should behave when there is no service available in your local zone. With this feature, you have the flexibility to configure the fallback behavior of your service, specifying the order in which it should attempt fallback options and defining different behaviors for instances located in various zones.
+
+- **`crossZone`** - (optional) allows to define behaviour when there is no healthy instances of the service. When not defined, cross zone traffic is disabled.
+  - **`failover`** - defines list of load balancing rules in order of priority. If a zone is not specified explicitly or by type `Any`, by default zone is excluded from the traffic.
```
slonka, line 52: It's a little bit convoluted, I'd go with:

```suggestion
  - **`failover`** - defines a list of load balancing rules in order of priority. If a zone is not specified explicitly by name or implicitly using the type `Any`/`AnyExcept` it is excluded from receiving traffic.
```

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -237,6 +269,239 @@ Apply the configuration with `kumactl apply -f [..]` or with the [HTTP API](/doc
 {% endtab %}
 {% endtabs %}
 
+{% if_version gte:2.5.x %}
+### Disable cross zone traffic and prioritize traffic the dataplanes on the same node and availability zone
+
+Requests to `backend` will be spread based on default weights and won't leave the local zone. 
+In this example, whenever user sends a request to `backend` service 90% of requests will arrive to the instance with the same value of `k8s.io/node` tag, 9% of requests to the instance with the same value as caller of tag `k8s.io/az`, and 1% to rest of instances.
```
slonka, line 276:
```suggestion
In this example, whenever a user sends a request to the `backend` service, 90% of the requests will arrive at the instance with the same value of the `k8s.io/node` tag, 9% of the requests will go to the instance with the same value as the caller of the `k8s.io/az` tag, and 1% will go to the rest of the instances.
```

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -237,6 +269,239 @@ Apply the configuration with `kumactl apply -f [..]` or with the [HTTP API](/doc
...
+Apply the configuration with `kumactl apply -f [..]` or with the [HTTP API](/docs/{{ page.version }}/reference/http-api).
+
+{% endtab %}
+{% endtabs %}
+
+### Prioritize traffic to dataplanes within the same datacenter and fallback cross zone in specific order
+
+Requests to backend will be distributed based on weights, with 99.9% of requests routed to data planes in the same datacenter, 0.001% to data planes in the same region, and the remainder to other local instances.
+
+When there are no healthy backends within the local zone, traffic from data planes in zones `us-1`, `us-2`, and `us-3` will only fall back to zones `us-1`, `us-2`, and `us-3`. If there are no healthy instances in these zones, requests from any instance will then fall back to `us-4`. If there are no healthy instances in `us-4`, the request will fail, as the last rule, by default, has type: `None`, meaning no fallback is allowed.
```
slonka, line 412: The "last rule" having type: None makes sense, but I think it's not mentioned anywhere else. Should we add it somewhere before giving examples?

Why isn't `eu` mentioned in this explanation even though it's defined [here](https://github.com/kumahq/kuma-website/pull/1506/files#diff-a9f8e0fd2b5222ab80b0bf763aefff4b9387802a05f2ad757de7a036815beb18R448)?

nitpick: 
```suggestion
When no healthy backends are available within the local zone, traffic from data planes in zones `us-1`, `us-2`, and `us-3` will only fall back to zones `us-1`, `us-2`, and `us-3`. If there are no healthy instances in these zones, requests from any instance will then fall back to us-4. If there are no healthy instances in us-4, the request will fail, as the last rule, by default, has a type of None, meaning no fallback is allowed.
```

slonka (changes requested):
Added some comments and nitpicks and one thing that I think is clearly wrong. 

Also, the egress part will be in a separate PR right?

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -237,6 +269,239 @@ Apply the configuration with `kumactl apply -f [..]` or with the [HTTP API](/doc
...
+                  type: Only
+                  zones: ["us-1"]
+              - to:
+                  type: AnyExcept
+                  zones: ["us-2", "us-3"]
+              - to:
+                  type: Any
+            failoverThreshold:
+              percentage: 25
+```
```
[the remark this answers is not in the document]
slonka: it auto convert to `k8s` right?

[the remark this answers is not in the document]
jakubdyszkiewicz: yes

app/_src/policies/meshloadbalancingstrategy.md
lukidzi: Changed

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -32,14 +32,14 @@ across all endpoints regardless of locality.
...
-  - **`affinityTags`** - list of tags and their weights based on which traffic is loadbalanced
-    - **`key`** - defines tag for which affinity is configured. The tag needs to be configured on the inbound of the service. In case of Kubernetes pod needs to have a tag and it should be propagated. On Universal user needs to define it on the inbound of the service. In case there so match, the dataplane match to the last group with the lowest weight.
+- **`localZone`** - (optional) allows to define load balancing priorities between dataplanes in the local zone. When not defined, traffic is distributed equally to all endpoints within the local zone.
+  - **`affinityTags`** - list of tags and their weights based on which traffic is load balanced
+    - **`key`** - defines tag for which affinity is configured. The tag needs to be configured on the inbound of the service. In case of Kubernetes, pod needs to have an annotations or a label. On Universal user needs to define it on the inbound of the service. In case there no match, the dataplane match to the last group with the lowest weight.
```
slonka, line 42: nitpick: missing "is"

```suggestion
    - **`key`** - defines tag for which affinity is configured. The tag needs to be configured on the inbound of the service. In case of Kubernetes, pod needs to have an annotations or a label. On Universal user needs to define it on the inbound of the service. In case there is no match, the dataplane match to the last group with the lowest weight.
```

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -61,6 +61,15 @@ Advanced locality-aware load balancing provides a powerful means of defining how
       - **`zones`** - names of zones
   - **`failoverThreshold.percentage`** - (optional) defines the percentage of live destination dataplane proxies below which load balancing to the next priority starts. Default: 50%
 
+#### Zone Egress support
+
+Using Zone Egress Proxy in multizone deployment poses certain limitations for this policy. When configuring `MeshLoadbalancingStrategy` with Zone Egress you can only use `Mesh` as a top level targetRef. This is because we don't differentiate requests that come to Zone Egress from different clients, yet. 
+
+Moreover, Zone Egress is a simple proxy that uses a long living L4 connection between Zone Ingress. Because of this, when new `MeshLoadbalancingStrategy` with locality awareness is configured, we won’t refresh connections and locality awareness will apply only to new connections.
```
slonka, line 68: > between Zone Ingress.

This does part is missing something, I think you need a plural after "between" or `and`.

```suggestion
Moreover, Zone Egress is a simple proxy that uses a long-living L4 connection to Zone Ingress. Consequently, when a new `MeshLoadbalancingStrategy` with locality awareness is configured, connections won’t be refreshed, and locality awareness will apply only to new connections.
```

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -61,6 +61,15 @@ Advanced locality-aware load balancing provides a powerful means of defining how
...
+
+Moreover, Zone Egress is a simple proxy that uses a long living L4 connection between Zone Ingress. Because of this, when new `MeshLoadbalancingStrategy` with locality awareness is configured, we won’t refresh connections and locality awareness will apply only to new connections. 
+
+Another thing you need to be aware of is that when you use the `MeshCircuitBreaker`'s outlier detection to keep track of healthy endpoints. Normally, you would use `MeshCircuitBreaker` to act on failures and trigger traffic redirect to the next priority level if the number of healthy endpoints fall below crossZone `failoverThreshold`. When you have a single instance of Zone Egress, all remote zones will be behind a single endpoint. Since `MeshCircuitBreaker` is configured on Data Plane Proxy, when one of the zones start responding with errors it will mark the whole Zone Egress as not healthy and won’t send traffic there even though there could be multiple zones with live endpoints.
```
slonka, line 70: I think the first sentence is missing the "conclusion" (?):

> Another thing you need to be aware of is * that when you use the `MeshCircuitBreaker`'s outlier detection to keep track of healthy endpoints.

where "*" - I think you need to specify what happens if you're using outlier detection. 

```suggestion
Another thing you need to be aware of is how outbound traffic behaves when you use the `MeshCircuitBreaker`'s outlier detection to keep track of healthy endpoints. Normally, you would use `MeshCircuitBreaker` to act on failures and trigger traffic redirect to the next priority level if the number of healthy endpoints fall below `crossZone.failoverThreshold`. When you have a single instance of Zone Egress, all remote zones will be behind a single endpoint. Since `MeshCircuitBreaker` is configured on Data Plane Proxy, when one of the zones start responding with errors it will mark the whole Zone Egress as not healthy and won’t send traffic there even though there could be multiple zones with live endpoints.
```

slonka: Also in general, should we mention that we intend to change this?

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -300,12 +302,13 @@ spec:
...
               - key: k8s.io/az
 ```
+{% endpolicy_yaml %}
 
-Apply the configuration with `kubectl apply -f [..]`.
+### Disable cross zone traffic and route to the local zone instances equally
 
-{% endtab %}
-{% tab local-zone-affinity-backend Universal %}
+In this example, when a user sends a request to the backend service, the request is routed equally to all instances in the local zone. If there are no instances in the local zone, the request will fail because there is no cross-zone traffic.
```
slonka, line 309: nitpick: "cross zone" in the title of subsection, but "cross-zone" in the contents, should we stick with one way? Amazon uses "cross-zone" - https://docs.aws.amazon.com/elasticloadbalancing/latest/userguide/how-elastic-load-balancing-works.html

slonka (changes requested):
Some changes needed. Also during test Friday we had this situation when we scaled down redis in one zone to 0 instances and then the traffic was completely not working even though we had failover defined. Switching address to `.mesh` helped. I think we should document this as well. Also should we document the pooling problem? When we only send 1 request concurrently then the connection for subsequent requests is always reused from the pool and it might not be load balanced across different backends (this is more of a demo scenario, not real prod environment, but might be helpful for people just trying this out).

lahabana:
Can we update this page:
https://kuma.io/docs/dev/policies/locality-aware/ ?

Let's make sure we say folks should stop using this and use our new more expressive feature instead

lahabana:
These titles are confusing:

image

It's only advanced because we used to have something before. I think now there should be a single title "laLB" and then subsections:

- default behaviour
- disabling cross-zone routing
- configuring laLB

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -30,6 +30,47 @@ Locality-aware load balancing is enabled by default unlike its predecessor [loca
...
+
+- **`localZone`** - (optional) allows to define load balancing priorities between dataplanes in the local zone. When not defined, traffic is distributed equally to all endpoints within the local zone.
+  - **`affinityTags`** - list of tags and their weights based on which traffic is load balanced
+    - **`key`** - defines tag for which affinity is configured. The tag needs to be configured on the inbound of the service. In case of Kubernetes, pod needs to have an annotations or a label. On Universal user needs to define it on the inbound of the service. In case there no match, the dataplane match to the last group with the lowest weight.
```
lahabana, line 42:
```suggestion
    - **`key`** - defines tag for which affinity is configured. The tag needs to be configured on the inbound of the service. In case of Kubernetes, pod needs to have a label. On Universal user needs to define it on the inbound of the service. In case there no match, the dataplane match to the last group with the lowest weight.
```

It doesn't work with annotations right?

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -30,6 +30,47 @@ Locality-aware load balancing is enabled by default unlike its predecessor [loca
...
 across all endpoints regardless of locality.
 
+{% if_version gte:2.5.x %}
+### Advanced LocalityAwareness
+Advanced locality-aware load balancing provides a more robust and straightforward method for balancing traffic within and across zones. This not only allows you to route traffic across zones when the local zone service is unhealthy but also enables you to define traffic prioritization within the local zone and set cross-zone fallback priorities.
+
+#### LocalZone
+Local zone routing allows you to define traffic routing rules within a local zone, prioritizing data planes based on tags and their associated weights. This enables you to allocate specific traffic percentages to data planes with particular tags within the local zone. If there are no healthy endpoints within the highest priority group, the next priority group takes precedence. Locality awareness within the local zone relies on tags within inbounds, so it's crucial to ensure that the tags used in the policy are defined for the service (Dataplane object on Universal, Deployment annotations on Kubernetes).
```
lahabana, line 38:
```suggestion
Local zone routing allows you to define traffic routing rules within a local zone, prioritizing data planes based on tags and their associated weights. This enables you to allocate specific traffic percentages to data planes with particular tags within the local zone. If there are no healthy endpoints within the highest priority group, the next priority group takes precedence. Locality awareness within the local zone relies on tags within inbounds, so it's crucial to ensure that the tags used in the policy are defined for the service (Dataplane object on Universal, PodTemplate labels on Kubernetes).
```

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -30,6 +30,47 @@ Locality-aware load balancing is enabled by default unlike its predecessor [loca
...
+
+- **`localZone`** - (optional) allows to define load balancing priorities between dataplanes in the local zone. When not defined, traffic is distributed equally to all endpoints within the local zone.
+  - **`affinityTags`** - list of tags and their weights based on which traffic is load balanced
+    - **`key`** - defines tag for which affinity is configured. The tag needs to be configured on the inbound of the service. In case of Kubernetes, pod needs to have an annotations or a label. On Universal user needs to define it on the inbound of the service. In case there no match, the dataplane match to the last group with the lowest weight.
+    - **`weight`** - (optional) weight of the tag used for load balancing. The bigger the weight the higher number of requests is routed to dataplanes with specific tag. By default we will adjust them so that 90% traffic goes to first tag, 9% to next, and 1% to third and so on.
+
+#### CrossZone
+{% warning %}
+Remember that cross-zone traffic requires [mTLS enabled](/docs/{{ page.version}}/policies/mutual-tls).
```
lahabana, line 47:
```suggestion
Remember that cross-zone traffic requires [mTLS to be enabled](/docs/{{ page.version}}/policies/mutual-tls).
```

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -30,6 +30,47 @@ Locality-aware load balancing is enabled by default unlike its predecessor [loca
...
+{% endwarning %}
+Advanced locality-aware load balancing provides a powerful means of defining how your service should behave when there is no service available in your local zone. With this feature, you have the flexibility to configure the fallback behavior of your service, specifying the order in which it should attempt fallback options and defining different behaviors for instances located in various zones.
```
lahabana, line 49:
```suggestion
Advanced locality-aware load balancing provides a powerful means of defining how your service should behave when there is no instances of your service available or they are in a degraded state in your local zone. With this feature, you have the flexibility to configure the fallback behavior of your service, specifying the order in which it should attempt fallback options and defining different behaviors for instances located in various zones.
```

I'm not sure it's the right wording but I think it's worth mentioning that if things degrade we do XZone

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -30,6 +30,47 @@ Locality-aware load balancing is enabled by default unlike its predecessor [loca
...
+
+- **`crossZone`** - (optional) allows to define behaviour when there is no healthy instances of the service. When not defined, cross zone traffic is disabled.
```
lahabana, line 51: I'm fine with this but it needs to be coherent with the `disabled` above

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -30,6 +30,47 @@ Locality-aware load balancing is enabled by default unlike its predecessor [loca
...
+  - **`failover`** - defines a list of load balancing rules in order of priority. If a zone is not specified explicitly by name or implicitly using the type `Any`/`AnyExcept` it is excluded from receiving traffic. By default, the last rule is always `None` which means, that there is no traffic to other zones after specified rules.
+    - **`from`** - (optional) defines the list of zones to which the rule applies. If not specified, rule is applied to all zones.
+      - **`zones`** - list of zones.
+    - **`to`** - defines to which zones the traffic should be load balanced.
+      - **`type`** - defines how target zones will be picked from available zones. Available options:
+        - **`Any`** - traffic will be load balanced to every available zone.
+        - **`Only`** - traffic will be load balanced only to zones specified in zones list.
+        - **`AnyExcept`** - traffic will be load balanced to every available zone except these specified in zones list.
```
lahabana, line 59:
```suggestion
        - **`AnyExcept`** - traffic will be load balanced to every available zone except those specified in zones list.
```

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -30,6 +30,47 @@ Locality-aware load balancing is enabled by default unlike its predecessor [loca
...
+        - **`None`** - traffic will not be load balanced to any zone.
+      - **`zones`** - names of zones
```
lahabana, line 61:
```suggestion
      - **`zones`** - list of zone names
```

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -30,6 +30,47 @@ Locality-aware load balancing is enabled by default unlike its predecessor [loca
...
+Remember that cross-zone traffic requires [mTLS enabled](/docs/{{ page.version}}/policies/mutual-tls).
+{% endwarning %}
+Advanced locality-aware load balancing provides a powerful means of defining how your service should behave when there is no service available in your local zone. With this feature, you have the flexibility to configure the fallback behavior of your service, specifying the order in which it should attempt fallback options and defining different behaviors for instances located in various zones.
+
+- **`crossZone`** - (optional) allows to define behaviour when there is no healthy instances of the service. When not defined, cross zone traffic is disabled.
+  - **`failover`** - defines a list of load balancing rules in order of priority. If a zone is not specified explicitly by name or implicitly using the type `Any`/`AnyExcept` it is excluded from receiving traffic. By default, the last rule is always `None` which means, that there is no traffic to other zones after specified rules.
+    - **`from`** - (optional) defines the list of zones to which the rule applies. If not specified, rule is applied to all zones.
+      - **`zones`** - list of zones.
```
lahabana, line 54:
```suggestion
      - **`zones`** - list of zone names.
```

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -30,6 +30,47 @@ Locality-aware load balancing is enabled by default unlike its predecessor [loca
...
+    - **`to`** - defines to which zones the traffic should be load balanced.
+      - **`type`** - defines how target zones will be picked from available zones. Available options:
+        - **`Any`** - traffic will be load balanced to every available zone.
+        - **`Only`** - traffic will be load balanced only to zones specified in zones list.
+        - **`AnyExcept`** - traffic will be load balanced to every available zone except these specified in zones list.
+        - **`None`** - traffic will not be load balanced to any zone.
+      - **`zones`** - names of zones
+  - **`failoverThreshold.percentage`** - (optional) defines the percentage of live destination dataplane proxies below which load balancing to the next priority starts. Default: 50%
```
lahabana, line 62: What's the format? Float? String?

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -30,6 +30,47 @@ Locality-aware load balancing is enabled by default unlike its predecessor [loca
...
+
+#### Zone Egress support
+
+Using Zone Egress Proxy in multizone deployment poses certain limitations for this policy. When configuring `MeshLoadbalancingStrategy` with Zone Egress you can only use `Mesh` as a top level targetRef. This is because we don't differentiate requests that come to Zone Egress from different clients, yet.
```
lahabana, line 66:
```suggestion
Using Zone Egress Proxy in multizone deployment poses certain limitations for this feature. When configuring `MeshLoadbalancingStrategy` with Zone Egress you can only use `Mesh` as a top level targetRef. This is because we don't differentiate requests that come to Zone Egress from different clients, yet. 
```

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -30,6 +30,47 @@ Locality-aware load balancing is enabled by default unlike its predecessor [loca
...
+        - **`AnyExcept`** - traffic will be load balanced to every available zone except these specified in zones list.
+        - **`None`** - traffic will not be load balanced to any zone.
+      - **`zones`** - names of zones
+  - **`failoverThreshold.percentage`** - (optional) defines the percentage of live destination dataplane proxies below which load balancing to the next priority starts. Default: 50%
+
+#### Zone Egress support
+
+Using Zone Egress Proxy in multizone deployment poses certain limitations for this policy. When configuring `MeshLoadbalancingStrategy` with Zone Egress you can only use `Mesh` as a top level targetRef. This is because we don't differentiate requests that come to Zone Egress from different clients, yet. 
+
+Moreover, Zone Egress is a simple proxy that uses a long-living L4 connection to Zone Ingress. Consequently, when a new `MeshLoadbalancingStrategy` with locality awareness is configured, connections won’t be refreshed, and locality awareness will apply only to new connections.
```
lahabana, line 68:
```suggestion
Moreover, Zone Egress is a simple proxy that uses long-lived L4 connection with each Zone Ingresses. Consequently, when a new `MeshLoadbalancingStrategy` with locality awareness is configured, connections won’t be refreshed, and locality awareness will apply only to new connections.
```

app/_src/policies/meshloadbalancingstrategy.md
Automaat: added info that this will be fixed in the future

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -237,6 +278,153 @@ Apply the configuration with `kumactl apply -f [..]` or with the [HTTP API](/doc
...
+spec:
+  targetRef:
+    kind: Mesh
+  to:
+    - targetRef:
+        kind: MeshService
+        name: backend
+      default:
+        localityAwareness:
+          localZone: {}
```
lahabana, line 346: Is this the same as not having a policy altogether?

lukidzi: No, it's like it works for `disabled`. If there is 
```
type: MeshLoadBalancingStrategy
name: local-zone-affinity-backend
mesh: mesh-1
spec:
  targetRef:
    kind: Mesh
  to:
    - targetRef:
        kind: MeshService
        name: backend
      default:
        localityAwareness: {}```

then `disabled: false`  and it's configured cross zone traffic with priority 1. We are checking if `localZone` or `crossZone` is configured to. configure the policy, which can be and empty object.

File: app/_src/policies/meshloadbalancingstrategy.md (hunk 17 above)
Comment (lukidzi) [reply app/_src/policies/meshloadbalancingstrategy.md:51]: If there is no `localZone` or `crossZone` section then `disabled` is used. If `localZone` or `crossZone` defined then new logic takes place.

File: app/_src/policies/meshloadbalancingstrategy.md (hunk 24 above)
Comment (lahabana) [reply app/_src/policies/meshloadbalancingstrategy.md:346]: Ok so this contradicts the at the top of this doc:

> Locality-aware load balancing is enabled by default unlike its predecessor [localityAwareLoadBalancing](https://deploy-preview-1506--kuma.netlify.app/docs/2.5.x/policies/locality-aware).

Comment (lahabana) [reply app/_src/policies/meshloadbalancingstrategy.md:346]: Btw: 

> When disabled requests are distributed across all endpoints regardless of locality

Is this still a thing? What does "disabled: true" do now?

File: app/_src/policies/meshloadbalancingstrategy.md [hunk 25]
```diff
@@ -30,6 +30,47 @@ Locality-aware load balancing is enabled by default unlike its predecessor [loca
...
+
+- **`localZone`** - (optional) allows to define load balancing priorities between dataplanes in the local zone. When not defined, traffic is distributed equally to all endpoints within the local zone.
+  - **`affinityTags`** - list of tags and their weights based on which traffic is load balanced
+    - **`key`** - defines tag for which affinity is configured. The tag needs to be configured on the inbound of the service. In case of Kubernetes, pod needs to have an annotations or a label. On Universal user needs to define it on the inbound of the service. In case there is no match, the dataplane match to the last group with the lowest weight.
```
lahabana, line 42:
```suggestion
    - **`key`** - defines tag for which affinity is configured. The tag needs to be configured on the inbound of the service. In case of Kubernetes, pod needs to have a label. On Universal user needs to define it on the inbound of the service. In case there is no match, the dataplane match to the last group with the lowest weight.
```

No annotations support.

app/_src/policies/meshloadbalancingstrategy.md
lahabana: Hmm it's odd. I though disabled flags were aimed as kill switches in new policies no?

lukidzi: If you define `disable: true` but you define `crossZone` then it would take `crossZone` rules.

lukidzi: We thought like these were 2 separate features, and that `disabled` usecase https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/031-locality-aware-load-balancing.md#load-balance-equally-to-local-and-zone

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -23,13 +23,63 @@ To learn more about the information in this table, see the [matching docs](/docs
...
 
 Locality-aware load balancing is enabled by default unlike its predecessor [localityAwareLoadBalancing](/docs/{{ page.version }}/policies/locality-aware).
 
 - **`disabled`** – (optional) allows to disable locality-aware load balancing. When disabled requests are distributed 
 across all endpoints regardless of locality.
 
+{% endif_version %}
+{% if_version gte:2.5.x %}
+### LocalityAwareness
+Locality-aware load balancing provides a more robust and straightforward method for balancing traffic within and across zones. This not only allows you to route traffic across zones when the local zone service is unhealthy but also enables you to define traffic prioritization within the local zone and set cross-zone fallback priorities.
```
lahabana, line 37:
```suggestion
Locality-aware load balancing provides robust and straightforward method for balancing traffic within and across zones. This not only allows you to route traffic across zones when the local zone service is unhealthy but also enables you to define traffic prioritization within the local zone and set cross-zone fallback priorities.
```

We should write docs in an intemporal tone it's easier to maintain

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -23,13 +23,63 @@ To learn more about the information in this table, see the [matching docs](/docs
...
+
+#### Default behaviour
+Locality-aware load balancing is enabled by default, unlike its predecessor [localityAwareLoadBalancing](/docs/{{ page.version }}/policies/locality-aware). Requests are distributed across all endpoints within the local zone first unless there are not enough healthy endpoints.
+
+#### Disabling cross zone routing
+It's possible to disable cross-zone traffic. In this case, all endpoints are treated equally.
```
lahabana, line 43: Can we add how to do this here?

lahabana: What you are actually disabling is localityAware LoadBalancing so this title and the first sentence is wrong. 

#### Disabling locality aware routing

If you do so, all endpoints regardless of their zone will be treated equally. To do this do:

```yaml
localityAwareness:
  disabled: true
```

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -23,13 +23,63 @@ To learn more about the information in this table, see the [matching docs](/docs
...
+
+#### Configuring LocalityAware Load Balancing
```
lahabana, line 45:
```suggestion
#### Configuring LocalityAware Load Balancing for traffic within the same zone
```

app/_src/policies/meshloadbalancingstrategy.md
```diff
@@ -23,13 +23,63 @@ To learn more about the information in this table, see the [matching docs](/docs
...
+
+- **`localZone`** - (optional) allows to define load balancing priorities between dataplanes in the local zone. When not defined, traffic is distributed equally to all endpoints within the local zone.
+  - **`affinityTags`** - list of tags and their weights based on which traffic is load balanced
+    - **`key`** - defines tag for which affinity is configured. The tag needs to be configured on the inbound of the service. In case of Kubernetes, pod needs to have a label. On Universal user needs to define it on the inbound of the service. If the tag is absent this entry is skipped.
+    - **`weight`** - (optional) weight of the tag used for load balancing. The bigger the weight the higher number of requests is routed to dataplanes with specific tag. By default we will adjust them so that 90% traffic goes to first tag, 9% to next, and 1% to third and so on.
+
```
lahabana, line 54: Add
```
#### Configuring LocalityAware Load Balancing for traffic across zones
```

app/_src/policies/locality-aware.md
```diff
@@ -2,6 +2,12 @@
 title: Locality-aware Load Balancing
 ---
 
+{% if_version gte:2.5.x %}
+{% warning %}
+If you are looking for more advanced configuration use [MeshLoadBalancingStrategy](/docs/{{ page.version }}/policies/meshloadbalancingstrategy).
```
lahabana, line 7:
```suggestion
This mode of doing locality aware load balancing is being replaced by [MeshLoadBalancingStrategy](/docs/{{ page.version }}/policies/meshloadbalancingstrategy) which is more powerful and flexible.
```

PR #1517, opened 2023-11-07
fix: traffic-metrics example

lahabana:
<Explain your change!>

### DIFF
--- a/app/_src/policies/traffic-metrics.md
+++ b/app/_src/policies/traffic-metrics.md
@@ -95,8 +95,6 @@ metrics:
   backends:
   - name: prometheus-1
     type: prometheus
-    conf:
-      skipMTLS: true # by default mTLS metrics are also protected by mTLS. Scraping metrics with mTLS without transparent proxy is not supported at the moment.
 ```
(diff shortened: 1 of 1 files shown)


app/_src/policies/traffic-metrics.md
```diff
@@ -54,7 +54,8 @@ spec:
     - name: prometheus-1
       type: prometheus
       conf:
-        skipMTLS: false
```
lukidzi, line 57: We should fix the example in line 102 for universal which configures tls disabled but here we create tls enabled.

app/_src/policies/traffic-metrics.md
```diff
@@ -54,7 +54,8 @@ spec:
...
+        tls:
+          mode: activeMTLSBackend
```
slonka, line 58: with this change the <= 2.3 version is the same as >= 2.4 version... it does not make sense for me

lahabana:
@slonka  and @lukidzi fixed. This doc is very hard to follow :)

lukidzi (approved):

PR #1551, opened 2023-12-11
ci(vale): try to reduce linter noise

lahabana:


michaelbeaumont (approved):
Is this for links? My suspicion is that vale is somehow just misconfigured because it's made for markdown, it can't be that it doesn't understand links, right?
But if it works it works.

michaelbeaumont:
@lahabana should we merge this?

PR #1574, opened 2024-01-15
docs(policy/VirtualOutbound): improvements in policy docs

bartsmykla:
- Fixes incorrectly formatted use cases list
- Adds Kubernetes version of one of the example policies, which was shown only in Universal format before
- Improves the additional requirements section

### DIFF
--- a/app/_src/policies/virtual-outbound.md
+++ b/app/_src/policies/virtual-outbound.md
@@ -6,10 +6,10 @@ This policy lets you customize hostnames and ports for communicating with data p
 
 Possible use cases are:
 
-1) Preserving hostnames when migrating to service mesh.
-2) Providing multiple hostnames for reaching the same service, for example when renaming or for usability.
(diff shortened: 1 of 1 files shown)


app/_src/policies/virtual-outbound.md
```diff
@@ -41,6 +41,33 @@ inbound:
 
 and a virtual outbound with this definition:
 
+{% tabs default-test-virtualoutbound useUrlFragment=false %}
+{% tab default-test-virtualoutbound Kubernetes %}
```
slonka, line 45: we have this `policy_yaml` function I think we should use: https://github.com/kumahq/kuma-website/blob/0b726a69960f8f5041a884902df58ea24a3604b0/jekyll-kuma-plugins/lib/jekyll/kuma-plugins/liquid/tags/policyyaml.rb#L1 - it takes `universal` format

slonka (changes requested):
One small thing to change

app/_src/policies/virtual-outbound.md
bartsmykla: it doesn't work on the legacy policies

PR #1609, opened 2024-01-25
feat(installer): script refresh/cloudsmith

curiositycasualty:
This change causes the installer script to source packages from Kong's new package hosting provider, Cloudsmith: [https://packages.konghq.com](https://packages.konghq.com)

Wherever possible, I tried to reuse the existing log messages such that documentation should not need to be re-written.
Features-wise, everything is the same except I also added `latest` as a possible input for `VERSION` (in addition to the existing `preview`).

Since the "release" package repos only contain 2.2.x+ and only support `linux`, `darwin` (kuma) and `linux`, `darwin`, and `windows` (kong-mesh), I initially thought I could drop the changes from: https://github.com/kumahq/kuma-website/pull/806/files (as discussed with @lahabana). But after some thought and some uploads of older packages to the Cloudsmith package repo "[kuma-legacy](https://cloudsmith.io/~kong/repos/kuma-legacy/packages)" (a rough analogue to Pulp's "mesh-alpine"), it seemed feasible with little effort, so I went ahead and kept it.

Did you sign your commit? [Instructions](https://github.com/kumahq/.github/blob/main/CONTRIBUTING.md#sign-your-commits) Yes

Have you read [Contributing guidelines](https://github.com/kumahq/.github/blob/main/CONTRIBUTING.md)? Yes

PR #1625, opened 2024-01-29
feat(installer): Source Packages from Cloudsmith

curiositycasualty:
This change causes the installer script to source packages from Kong's new package hosting provider, Cloudsmith: [https://packages.konghq.com](https://packages.konghq.com)

Wherever possible, I tried to reuse the existing log messages such that documentation should not need to be re-written.
Features-wise, everything is the same except I also added `latest` as a possible input for `VERSION` (in addition to the existing `preview`).

Since the "release" package repos only contain 2.2.x+ and only support `linux`, `darwin` (kuma) and `linux`, `darwin`, and `windows` (kong-mesh), I initially thought I could drop the changes from: https://github.com/kumahq/kuma-website/pull/806/files (as discussed with @lahabana). But after some thought and some uploads of older packages to the Cloudsmith package repo "[kuma-legacy](https://cloudsmith.io/~kong/repos/kuma-legacy/packages)" (a rough analogue to Pulp's "mesh-alpine"), it seemed feasible with little effort, so I went ahead and kept it.

Replaces: https://github.com/kumahq/kuma-website/pull/1609

[KAG-2257](https://konghq.atlassian.net/browse/KAG-2257)


.github/workflows/ci.yml
```diff
@@ -43,10 +44,37 @@ jobs:
...
+            --exclude 'https://github.com/spiffe/spiffe/blob/main/standards/X509-SVID.md.*' \
+            --exclude ${URL}/docs/1. \
+            --exclude 127.0.0.1 \
+            --exclude https://cloudsmith.io/~kong/repos \
+            --exclude https://linux.die.net \
+            --exclude https://packages.konghq.com \
+            --exclude https://twitter.com \
+            --max-connections-per-host=8 \
+            --rate-limit 50 \
+            --timeout 60
```
lahabana, line 61: Thx!

app/installer.sh
```diff
@@ -18,210 +18,261 @@
...
+    err "Required tools cannot be found: ${missing}."
   fi
 
-  if [ -z "$VERSION" ]; then
-    printf "ERROR\tUnable to fetch latest %s version because of a problem with %s.\n" "$PRODUCT_NAME" "$PRODUCT_NAME"
-    exit 1
+  if [ "$VERSION" = 'latest' ]; then
+    log "Fetching latest ${PRODUCT_NAME} version.."
+
+    LATEST_VERSION="$(curl -f -sL 'https://kuma.io/latest_version')"
```
lahabana, line 99: How do you expect this to work for KongMesh?

lahabana: I think you dropped the LATEST_VERSION url that we can change in the case of entreprise see: https://github.com/Kong/docs.konghq.com/blob/f31e8418de8657fd13182be828e5caab0ca81fcf/app/mesh/installer.sh#L24

lahabana: (worth actually checking this invocation locally :)

curiositycasualty: Ah, you're right. I didn't realize it was being passed into the script externally like this.
I assumed I could just set `VERSION` to the correct version if unsupplied.
Good catch, I'll toss in a fix for this.

.github/workflows/ci.yml
```diff
@@ -84,17 +117,36 @@ jobs:
...
 
           # Set an output parameter `uname` for use in subsequent steps
           run: |
-            apt-get update && apt-get install -y file curl tar gzip
+            apt-get update && apt-get install -y tar gzip curl
             sh app/installer.sh
             file ./kuma-*/bin/kumactl | grep aarch64 &> /dev/null
 
+  installer-sh-mesh:
+    name: Test installer.sh for Kong Mesh
```
slonka, line 134: we usually don't test anything related to KM in `kumahq`. Should this be in https://github.com/Kong/docs.konghq.com ?

app/installer.sh
```diff
@@ -18,210 +18,262 @@
...
+    if ! command -v $tool >/dev/null 2>&1; then
+      missing="${missing:+${missing} }${tool}"
+    fi
+  done
 
-  if ! VERSION=$(curl -s $LATEST_VERSION); then
-    printf "ERROR\tUnable to fetch latest %s version.\n" "$PRODUCT_NAME"
-    exit 1
+  if [ -n "$missing" ]; then
+    err "Required tools cannot be found: ${missing}."
```
slonka, line 94: great addition, thanks ❤️

slonka (commented):
Overall looks good, not 👍 yet because not sure if we want to test KM here as well.

.github/workflows/ci.yml
lahabana: I think it's fine. If anyone is against it (because it's against some CNCF policy we'll remove it)

slonka (approved):

curiositycasualty:
Restored `file` to the workflow 😔
I expect this to pass when approved/ran.

lahabana (approved):
It's so much quicker too ❤️

johnharris85:
LGTM

PR #1631, opened 2024-01-30
Fix auth header s/authentication/authorization

magec:
I was blindly copying http headers from reference and the server kept saying 'Permission denied'
```
{
 "type": "/std-errors",
 "status": 403,
 "title": "Access Denied",
 "detail": "action not allowed",
 "details": "action not allowed"
}
```
Then I realized is `authorization` not `authentication` (by reverse engeniering kumactl XD), change it and worked well. This change updates the doc.


lahabana (approved):

lahabana:
Thx for the contrib!

PR #1641, opened 2024-02-07
fix(installer): set VERSION to the commit hash

lobkovilya:
When running a command:
```
curl https://kuma.io/installer.sh | VERSION=0.0.0-preview.v97af5da5e sh -
```
it attempts to use the URL (and fails):
```
https://packages.konghq.com/public/kuma-binaries-preview/raw/names/kuma-darwin-arm64/versions/0.0.0-preview.v97af5da5e/kuma-0.0.0-preview.v97af5da5e-darwin-arm64.tar.gz
```
the correct URL is (difference after the `/versions/`):
```
https://packages.konghq.com/public/kuma-binaries-preview/raw/names/kuma-darwin-arm64/versions/97af5da5e/kuma-0.0.0-preview.v97af5da5e-darwin-arm64.tar.gz
```

### DIFF
--- a/app/installer.sh
+++ b/app/installer.sh
@@ -212,6 +212,7 @@ EOF
 
   if echo "$FILENAME_VERSION" | grep -qs -E 'preview|0.0.0'; then
     URL_REPO="${REPO_REPO}-binaries-preview"
+    VERSION="${FILENAME_VERSION#*0.0.0-preview.v}"
   else
 
     # 2.1.x or lower
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,8 +93,6 @@ jobs:
           - ubuntu:22.04
     container:
       image: ${{ matrix.os }}
-    env:
-      VERBOSE: ${{ runner.debug }}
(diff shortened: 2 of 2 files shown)


app/installer.sh
```diff
@@ -212,6 +212,7 @@ EOF
 
   if echo "$FILENAME_VERSION" | grep -qs -E 'preview|0.0.0'; then
     URL_REPO="${REPO_REPO}-binaries-preview"
+    VERSION="${FILENAME_VERSION#*0.0.0-preview.v}"
```
slonka, line 215: can we add a case for latest / specific commit here https://github.com/kumahq/kuma-website/blob/9948b52a52d83b0f4b7ce7800229b3dda25e2060/.github/workflows/ci.yml#L105 ?

slonka (commented):
Would really help us avoid breaking installer for different modes

app/installer.sh
slonka: and we could get latest version (to check specific version) from https://kuma.io/latest_version

PR #1650, opened 2024-02-13
fix(installer.sh): --print-version with preview versions

michaelbeaumont:
Behavior was changed at some point, [breaking kong-mesh-ecs](https://github.com/Kong/kong-mesh-ecs/actions/runs/7881728764/job/21505818554#step:6:1).

### DIFF
--- a/app/installer.sh
+++ b/app/installer.sh
@@ -177,7 +177,7 @@ main() {
   if [ $# -gt 0 ]; then
     case $1 in
     --print-version)
-      echo "$VERSION"
+      echo "${FILENAME_VERSION}"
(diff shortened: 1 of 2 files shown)


slonka (approved):

app/installer.sh
```diff
@@ -177,7 +177,7 @@ main() {
   if [ $# -gt 0 ]; then
     case $1 in
     --print-version)
-      echo "$VERSION"
+      echo "${FILENAME_VERSION}"
```
slonka, line 180: should we also cover this in our tests?

PR #1654, opened 2024-02-14
feat(guides): add kong delegated gateway guide

lahabana:
<Explain your change!>

### DIFF
--- a/app/_src/guides/gateway-delegated.md
+++ b/app/_src/guides/gateway-delegated.md
@@ -1,0 +1,197 @@
+---
+title: Use Kong as a delegated Gateway 
+---
+
+To get traffic from outside your mesh inside it (North/South) with {{site.mesh_product_name}} you can use 
+a delegated gateway.
+
+In the [quickstart](/docs/{{ page.version }}/quickstart/kubernetes-demo/), traffic was only able to get in the mesh by port-forwarding to an instance of an app
+inside the mesh.
+In production, you typically set up a gateway to receive traffic external to the mesh.
+In this guide you will add Kong as a [delegated gateway](/docs/{{ page.version }}/using-mesh/managing-ingress-traffic/delegated/) in front of the demo-app service and expose it publicly.
+
+{% mermaid %}
+---
+title: service graph of the demo app with a Kong gateway on front
+---
+flowchart LR
+  subgraph Kong Gateway 
+    gw0(/ :80)
+  end
+  demo-app(demo-app :5000)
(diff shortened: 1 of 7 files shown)


app/_src/guides/gateway-delegated.md
```diff
@@ -0,0 +1,197 @@
...
+  - matches:
+    - path:
+        type: PathPrefix
+        value: /
+    backendRefs:
+    - name: demo-app
+      kind: Service
+      port: 5000 
+" | kubectl apply -f -
+```
```
johncowen, line 118: It might be worth calling out that this is adding `HTTPRoute` to kong and not kuma (kinda 1% guessing here, but I think this is the case seeing as we have delegated all this to the 3rd party gateway). I had a double take for a second and was like "oooooooh, its not a `MeshHTTPRoute` its a `HTTPRoute` that must be a kong gateway thing". You kinda hint at this when I read it back again, but might be worth adding a little more?

Also, at first I thought there was a mistake in the code here as the `parentRef` uses the `default` namespace not the `kong` one as you mention. This might just be my misunderstanding, so this might be a query just for me, but also might be worth explaining that in the text, not sure up to you.

app/_src/guides/gateway-delegated.md
```diff
@@ -0,0 +1,197 @@
...
+}%
+```
+
+## Add a route to our `demo-app`
+
+Patch our gateway to allow routes in any namespace:
+
+```shell
+kubectl patch --type=json gateways.gateway.networking.k8s.io kong -p='[{"op":"replace","path": "/spec/listeners/0/allowedRoutes/namespaces/from","value":"All"}]'
+```
```
johncowen, line 94: Mostly for personal understanding (but might also be worth adding something), I wondered:

1. If we had to do this just because we are using kong i.e. is this necessary "because kong" or "because kuma"? Guessing its "because kong"?
2. Pretty sure this is allowing kong `HTTPRoute`s in any namespace, might be worth this is allowing kong routes not kuma ones.

I think as I'm writing this I'm kinda answering my own questions in my head and things are making more sense. I think generally it might be worth adding and underlining the things that are different specifically because this is a delegated 3rd party gateway, umm I suppose the "why we are doing these bits".

lahabana: It's actually GatewayAPI

johncowen: Oooh I see interesting!

PR #1680, opened 2024-03-06
docs(production): clarify port number in kube service

jakubdyszkiewicz:
Clarify that admission server on 5443 is exposed on 443


cbugneac-nex:
For more context, here is Kume 2.6.1 CP service definition:
```
$ kubectl get svc kuma-control-plane -o yaml
...
spec:
  ports:
...
  - appProtocol: https
    name: https-admission-server
    port: 443
    protocol: TCP
    targetPort: 5443
...
```

Looks like the service is listening on port 443 but forwarding to pod port 5443.

lahabana (approved):

cbugneac-nex:
Thanks @lahabana

PR #1710, opened 2024-04-05
fix(installer.sh): make VERSION=preview more reliable

lahabana:
Before we were using some graphQL black magic.
Now we use `gh run list` which enables to very strictly filter on successful workflow run

This should avoid failures in CI due to build-test-distribute waiting to be run

### DIFF
--- a/app/installer.sh
+++ b/app/installer.sh
@@ -110,7 +110,7 @@ main() {
     fi
 
     log "Fetching latest preview commit.."
-    commit=$(gh run list --repo "${REPO}" --branch "${BRANCH}" --workflow build-test-distribute -e push -s success --json headSha --jq '. | first | .headSha[0:9]')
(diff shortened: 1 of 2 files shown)


slonka (approved):

app/installer.sh
```diff
@@ -112,63 +109,12 @@ main() {
...
-    for count in 5 10 25 50; do
-      commit="$(inner "$count")"
-
-      if echo "$commit" | grep -qs -E '[a-z0-9]{9,}'; then
-        break
-      fi
-    done
-
+    log "Fetching latest preview commit.."
+    commit=$(gh run list --repo "${REPO}" --branch "${BRANCH}" --workflow build-test-distribute -L 200 --json event,headSha,conclusion --jq '[.[] | select(.conclusion == "success" and .event == "push")] | first | .headSha[0:9]')
```
slonka, line 113: Seems like `--branch` is not available: https://github.com/kumahq/kuma-website/actions/runs/8570497791/job/23488536534?pr=1710#step:6:15

lahabana: Ha nice! Thx for digging in!

lahabana: Opened: https://github.com/kumahq/kuma-website/issues/1711 to track this

PR #1730, opened 2024-04-12
docs(policies): improve otel metrics guide

Automaat:
small improvement to make guide smoother for users


app/_src/guides/otel-metrics.md
```diff
@@ -107,19 +108,30 @@ We need to update `prometheus-server` ConfigMap and add `scrape_configs` entry:
     - targets: ["opentelemetry-collector.mesh-observability.svc:8889"]
 ```
 
-Prometheus will automatically pick up this config and start scraping OpenTelemetry collector.
+Prometheus will automatically pick up this config and start scraping OpenTelemetry collector. To check if config was applied properly
+you can go to Prometheus GUI:
+
+```shell
+kubectl port-forward svc/prometheus-server -n mesh-observability 9090:80
```
slonka, line 115: are we sure it's listening on 80? I Thought in k9s I always do `9090:9090` 🤔

slonka (approved):

app/_src/guides/otel-metrics.md
Automaat: service is listening on 80. I checked it before I've changed the docs

PR #1733, opened 2024-04-12
feat(installer): use cloudsmith instead of gh to fetch preview version

slonka:
`gh` requires a valid token - in this PR I exchange it for cloudsmith which does not and now people who don't have a valid token can try out the preview version

### DIFF
--- a/app/installer.sh
+++ b/app/installer.sh
@@ -105,12 +105,12 @@ main() {
 
   if [ "$VERSION" = 'preview' ]; then
 
-    if ! command -v gh >/dev/null 2>&1; then
-      err "Must have github's gh CLI installed to install a preview version."
+    if ! command -v jq >/dev/null 2>&1; then
+      err "Must have jq installed to install a preview version."
     fi
 
     log "Fetching latest preview commit.."
-    commit=$(gh run list --repo "${REPO}" --branch "${BRANCH}" --workflow build-test-distribute -L 200 --json event,headSha,conclusion --jq '[.[] | select(.conclusion == "success" and .event == "push")] | first | .headSha[0:9]')
+    commit=$(curl -s --request GET --url 'https://api.cloudsmith.io/v1/packages/kong/kong-mesh-binaries-preview/?page=1&page_size=1&query=filename%3A0.0.0-preview&sort=-date' --header 'accept: application/json' | jq -r '.[0].version')
     if ! echo "$commit" | grep -qs -E '[a-z0-9]{9,}'; then
       err "Failed to find suitable preview commit (${count} commits checked)."
     fi
--- a/app/installer.sh
+++ b/app/installer.sh
@@ -110,7 +110,7 @@ main() {
     fi
 
     log "Fetching latest preview commit.."
-    commit=$(curl -s --request GET --url "https://api.cloudsmith.io/v1/packages/kong/${PRODUCT_NAME}-binaries-preview/?page=1&page_size=1&query=filename%3A0.0.0-preview&sort=-date" --header 'accept: application/json' | jq -r '.[0].version')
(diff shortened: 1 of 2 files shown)


app/installer.sh
```diff
@@ -105,12 +105,12 @@ main() {
...
 
-    if ! command -v gh >/dev/null 2>&1; then
-      err "Must have github's gh CLI installed to install a preview version."
+    if ! command -v jq >/dev/null 2>&1; then
+      err "Must have jq installed to install a preview version."
     fi
 
     log "Fetching latest preview commit.."
-    commit=$(gh run list --repo "${REPO}" --branch "${BRANCH}" --workflow build-test-distribute -L 200 --json event,headSha,conclusion --jq '[.[] | select(.conclusion == "success" and .event == "push")] | first | .headSha[0:9]')
+    commit=$(curl -s --request GET --url "https://api.cloudsmith.io/v1/packages/kong/${PRODUCT_NAME}-binaries-preview/?page=1&page_size=1&query=filename%3A0.0.0-preview&sort=-date" --header 'accept: application/json' | jq -r '.[0].version')
```
lahabana, line 113: `curl https://deploy-preview-1733--kuma.netlify.app/installer.sh | PRODUCT_NAME=kuma VERSION=preview sh -`

Works but `curl https://deploy-preview-1733--kuma.netlify.app/installer.sh | VERSION=preview sh -` fails with:

```
curl https://deploy-preview-1733--kuma.netlify.app/installer.sh | VERSION=preview sh -
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  6571  100  6571    0     0  21904      0 --:--:-- --:--:-- --:--:-- 21903
INFO    Welcome to the Kuma automated download!
INFO    Fetching latest preview commit..
jq: error (at <stdin>:0): Cannot index object with number
```

lahabana: Product name is `Kuma` vs `kuma` this likely explains the issue

michaelbeaumont (changes requested):
So my objection to this was that this changes the meaning of `preview` from "successful CI" to "build exists." At one point we had builds even if the e2e tests failed. This seems to have been lost at some point during CI refactors.
So my objection is that changing this logic prevents us from adding that back, which I think we should have back.

app/installer.sh
slonka: fixed

lahabana:
The end story is why do we have previews in this script?

My original idea when I added this was: being able for external users to use a trusted build to check upcoming features, validate fixes... For this having `VERSION=preview` be something we think works (e2e pass) would be useful.
@michaelbeaumont definition of what this script should return seems different as: something for the dev team to debug which I think is different and orthogonal to this.

For @michaelbeaumont 's usecase you can always open a PR with e2e tests skipped to get something equivalent (or just build locally). I think using cloudsmith solves a bunch of access issues which I think is meaningful (external users able to use `VERSION=preview` is a use case).
If we want to do this real difference I think what should happen is build, push to a prerelease area (this includes docker images) then once validated move to the actual release area. In that case you could use cloudsmith (and docker) for both and even have `latest` docker tags.

michaelbeaumont:
> @michaelbeaumont definition of what this script should return seems different as: something for the dev team to debug which I think is different and orthogonal to this.

No. it's for allowing users to run KM independent of whether e2e tests complete. we had multiple cases where customers were trying to test out `master` builds but e2e flakes stopped this from happening.

michaelbeaumont (approved):
Approved, under the caveat that `VERSION=preview` now no longer has the same guarantees about stability. In particular, it may include builds that don't pass e2e tests and it may include non default branch builds.

Issue #1725, opened 2024-04-11
Check examples and tables to remove usages of `spec.from[].targetRef.kind: MeshService`


lahabana:
### What happened?

Since: https://github.com/kumahq/kuma/issues/9258 landed in 2.7 we should make sure docs don't advertise a deprecated use case.

kleinfreund:
The examples here need updating:

- https://kuma.io/docs/dev/policies/meshfaultinjection/#examples

PR #1736, opened 2024-04-14
fix(docs): update tables and examples as `from.kind: MeshService` is deprecated

lobkovilya:
Fix https://github.com/kumahq/kuma-website/issues/1725


app/_src/guides/gateway-builtin.md
```diff
@@ -204,8 +204,9 @@ spec:
     name: demo-app_kuma-demo_svc_5000
   from:
     - targetRef:
-        kind: MeshService
-        name: edge-gateway 
+        kind: MeshSubset
```
michaelbeaumont, line 207: Makes me wonder whether we should support `MeshGateway` as `from`, it doesn't have the same problems as `MeshService`

michaelbeaumont (approved):

Automaat (approved):

app/_src/guides/gateway-builtin.md
lobkovilya: Yeah, but I think at this moment there is an odd behaviour with multiple selectors https://github.com/kumahq/kuma/issues/9185#issuecomment-1973363863

Issue #1719, opened 2024-04-10
Update the Gateway API docs as it can be used in multi-zone now


bartsmykla:
### Description

[Gateway API docs](https://kuma.io/docs/dev/using-mesh/managing-ingress-traffic/gateway-api/) should be updated to reflect the fact that Gateway API resources can be used in multi-zone environments now.

PR #1738, opened 2024-04-15
docs(gatewayapi): fix and adjust gateway api documentation

bartsmykla:
Rendered, changed page: https://deploy-preview-1738--kuma.netlify.app/docs/2.7.x/using-mesh/managing-ingress-traffic/gateway-api/

Closes https://github.com/kumahq/kuma-website/issues/1719

- change CRDs installation step
  - Gateway API doesn't need to be installed in Experimental
    Channel anymore, as we are not supporting any of experimental
    features yet, so I changed the instruction to reflect that.
  - Admission Webhook doesn't come with CRDs anymore, so we can
    safely stop mentioning it
- update `apiVersion` of GAPI resources to `v1`
  Gateway API is GA now and we can change the version of resources
  from `v1beta1` to `v1`
- split `kubectl get ...` commands to separate blocks
  Now you can copy commands when want to follow instructions to the
  letter easier.
- remove unnecessary line from yaml block
- reformat `{% capture gateway %}`
  By using `{% capture ... %}` we can make sure rendered content
  will be correctly indented, but so far we unnecessarily put
  too much content there, so it caused confusion. Changed 
- don't render `MeshGatewayRoute`s in How it works for `>= 2.7.0`
  From 2.7.0 we are internally not generating `MeshGatewayRoute`s,
  so there is no need of mentioning it.
-  show no multi-zone for GAPI warning only for `<= 2.6.x`
- fix warning about GAMMA being experimental
  - GAMMA is GA now so the warning message was updated
  - instead of hardcoded `Kuma` in the warning we should use
    `{{site.mesh_product_name}}`
- adjust docs after promoting GAPI to GA
  Our Gateway API implementation is being promoted to GA in 2.7.0, so all warnings related to it are wrapped with `if_version` blocks.
- change code blocks to `yaml`
  To be consistent with the rest of the page and to make it easier for people which want to copy the resource to file I changed some of the code blocks from `sh` with instructions `echo "..." | kubectl apply -f-` to be just `yaml`s with pure manifests.
- update section about multi-zone deployments
  Gateway API is currently allowed with two limitations.

### DIFF
--- a/app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
+++ b/app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
@@ -17,7 +17,7 @@ Gateway API [`Gateways`](https://gateway-api.sigs.k8s.io/api-types/gateway/) are
 
 1. Install the Gateway API CRDs.
 
-   The Gateway API CRDs aren't available in Kubernetes by default yet. You must first [install the _experimental release_ that includes those CRDs as well as the admission webhook](https://gateway-api.sigs.k8s.io/guides/#install-experimental-channel).
+   Kubernetes doesn't include Gateway API functionality by default. To use it, you'll need to install the CRDs first. Instructions: [Installing Gateway API Standard Channel](https://gateway-api.sigs.k8s.io/guides/#install-standard-channel).
 
 2. Enable Gateway API support.
 
--- a/app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
+++ b/app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
@@ -198,13 +198,13 @@ metadata:
 spec:
   gatewayClassName: kuma
   listeners:
-    - name: proxy
-      port: 8080
-      hostname: test.kuma.io
-      protocol: HTTPS
-      tls:
-        certificateRefs:
-          - name: secret-tls
+  - name: proxy
+    port: 8080
+    hostname: test.kuma.io
+    protocol: HTTPS
+    tls:
+      certificateRefs:
+      - name: secret-tls
 ```
 
 Under the hood, {{site.mesh_product_name}} CP copies the `Secret` to `{{site.mesh_namespace}}` namespace and converts it to {% if_version lte:2.1.x %}[{{site.mesh_product_name}} secret](/docs/{{ page.version }}/security/secrets){% endif_version %}{% if_version gte:2.2.x %}[{{site.mesh_product_name}} secret](/docs/{{ page.version }}/production/secure-deployment/secrets/){% endif_version %}.
@@ -279,9 +279,9 @@ metadata:
 spec:
   gatewayClassName: kuma-cross-mesh
   listeners:
-    - name: proxy
-      port: 8080
-      protocol: HTTP
+  - name: proxy
+    port: 8080
+    protocol: HTTP
 ```
 
 ### Multi-zone Deployments
@@ -345,15 +345,15 @@ metadata:
   namespace: kuma-demo
 spec:
   parentRefs:
-    - name: demo-app
-      port: 5000
-      kind: Service
+  - name: demo-app
+    port: 5000
+    kind: Service
   rules:
-    - backendRefs:
-        - name: demo-app-v1
-          port: 5000
-        - name: demo-app-v2
-          port: 5000
+  - backendRefs:
+    - name: demo-app-v1
+      port: 5000
+    - name: demo-app-v2
+      port: 5000
 ```
 
 The namespace of the `HTTPRoute` is key. If the route's namespace and the
--- a/app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
+++ b/app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
@@ -19,7 +19,7 @@ Gateway API [`Gateways`](https://gateway-api.sigs.k8s.io/api-types/gateway/) are
 
 1. Install the Gateway API CRDs.
 
-   Kubernetes doesn't include Gateway API CRDs by default. To install them follow the [Installing Gateway API Standard Channel](https://gateway-api.sigs.k8s.io/guides/#install-standard-channel).
+   Kubernetes doesn't include Gateway API CRDs, [install them from the standard channel CRD bundle](https://gateway-api.sigs.k8s.io/guides/#install-standard-channel).
 
 2. Enable Gateway API support.
 
@@ -27,7 +27,7 @@ Gateway API [`Gateways`](https://gateway-api.sigs.k8s.io/api-types/gateway/) are
    - With Helm, use the `{{site.set_flag_values_prefix}}experimental.gatewayAPI=true` value.
 {% endif_version %}
 {% if_version gte:2.7.x %}
-Kubernetes doesn't include Gateway API CRDs by default. To install them follow the [Installing Gateway API Standard Channel](https://gateway-api.sigs.k8s.io/guides/#install-standard-channel).
+Kubernetes doesn't include Gateway API CRDs, [install them from the standard channel CRD bundle](https://gateway-api.sigs.k8s.io/guides/#install-standard-channel).
 {% endif_version %}
 
 {% if_version lte:2.2.x %}
@@ -294,7 +294,7 @@ Gateway API isn't supported with multi-zone deployments, use {{site.mesh_product
 The Gateway API supports multi-zone deployments, but with some limitations:
 
 - Gateway API resources like `Gateway`, `ReferenceGrant`, and `HTTPRoute` must be created in non-global zones. They are kubernetes resources and therefore are not known to {{ site.mesh_product_name }}.
-- Only services deployed within the same Kubernetes cluster, such as the `HTTPRoute`, can be referenced as the `backendRef`.
+- Only services deployed within the same Kubernetes cluster, such as the `HTTPRoute`, can be referenced via `backendRef`.
 
 {% capture backendref-limitation %}
 {% tip %}
@@ -402,11 +402,11 @@ flowchart TD
 {% endmermaid %}
 {% endcapture %}
 
-   - If in `k8s-cluster-1` you create an `HTTPRoute` with a `backendRef` targeting the `backend` service, it will only route traffic to the `backend` service in `k8s-cluster-1`.
+   - If you create an `HTTPRoute` with a `backendRef` targeting the `backend` service in `k8s-cluster-1`, it will only route traffic to the `backend` service in `k8s-cluster-1`.
      
      {{ gapi_multizone_limitation_1 | indent }}
 
-   - Similarly, if in `k8s-cluster-1` you create an `HTTPRoute` with a `backendRef` pointing to the `db` service, it will result in a `HTTPRoute` with a `ResolvedRefs` status condition of `BackendNotFound` because service `db` is not present in `k8s-cluster-1`.
+   - Similarly, if you create an `HTTPRoute` with a `backendRef` pointing to the `db` service in `k8s-cluster-1`, it will result in a `HTTPRoute` with a `ResolvedRefs` status condition of `BackendNotFound` because service `db` is not present in `k8s-cluster-1`.
 
      {{ gapi_multizone_limitation_2 | indent }}
 
--- a/app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
+++ b/app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
@@ -305,7 +305,7 @@ This is a temporary limitation, which will be lifted when [work on allowing targ
 
 Here's an example scenario that describes how you could configure multi-zone deployments with the Gateway API. In this example, you have the following resources:
 * Two zones (`zone-1` and `zone-2`) in separate Kubernetes clusters
-*  Two services:
+* Two services:
   * A service named `backend` deployed in each zone 
   * A service named `db` deployed only in `zone-1`
 
--- a/app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
+++ b/app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
@@ -304,10 +304,11 @@ This is a temporary limitation, which will be lifted when [work on allowing targ
 {{ backendref-limitation | indent }}
 
 Here's an example scenario that describes how you could configure multi-zone deployments with the Gateway API. In this example, you have the following resources:
-* Two zones (`zone-1` and `zone-2`) in separate Kubernetes clusters
-* Two services:
-  * A service named `backend` deployed in each zone 
-  * A service named `db` deployed only in `zone-1`
+
+- Two zones (`zone-1` and `zone-2`) in separate Kubernetes clusters
+- Two services:
+  - A service named `backend` deployed in each zone 
+  - A service named `db` deployed only in `zone-1`
 
 If you deploy multi-zone with Gateway API, the following will occur:
    
--- a/app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
+++ b/app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
@@ -68,53 +68,52 @@ spec:
   controllerName: gateways.kuma.io/controller
 " | kubectl apply -f -
 ```
-
 {% endtab %}
 {% endtabs %}
-
-```sh
-echo "apiVersion: gateway.networking.k8s.io/v1
-kind: Gateway
-metadata:
-  name: kuma
-  namespace: kuma-demo
-spec:
-  gatewayClassName: kuma
-  listeners:
-  - name: proxy
-    port: 8080
-    protocol: HTTP
-" | kubectl apply -f -
-```
-
-When a user applies a `Gateway` resource, {{site.mesh_product_name}} automatically creates a `Deployment` of built-in gateways with a corresponding `Service`.
-
-```sh
-kubectl get pods -n kuma-demo
-```
-
-```
-NAME                          READY   STATUS    RESTARTS   AGE
-redis-59c9d56fc-6gcbc         2/2     Running   0          2m8s
-demo-app-5845d6447b-v7npw     2/2     Running   0          2m8s
-kuma-4j6wr-58998b5576-25wl6   1/1     Running   0          30s
-```
-
-```sh
-kubectl get svc -n kuma-demo
-```
-
-```
-NAME         TYPE           CLUSTER-IP      EXTERNAL-IP   PORT(S)          AGE
-redis        ClusterIP      10.43.223.223   <none>        6379/TCP         3m27s
-demo-app     ClusterIP      10.43.216.203   <none>        5000/TCP         3m27s
-kuma-pfh4s   LoadBalancer   10.43.122.93    172.20.0.3    8080:30627/TCP   87s
-```
-
 {% endcapture %}
 
 {{ gateway | indent }}
-The `Gateway` is now accessible using the external address `172.20.0.3:8080`.
+
+   ```sh
+   echo "apiVersion: gateway.networking.k8s.io/v1
+   kind: Gateway
+   metadata:
+     name: kuma
+     namespace: kuma-demo
+   spec:
+     gatewayClassName: kuma
+     listeners:
+     - name: proxy
+       port: 8080
+       protocol: HTTP
+   " | kubectl apply -f -
+   ```
+   
+   When a user applies a `Gateway` resource, {{site.mesh_product_name}} automatically creates a `Deployment` of built-in gateways with a corresponding `Service`.
+   
+   ```sh
+   kubectl get pods -n kuma-demo
+   ```
+   
+   ```
+   NAME                          READY   STATUS    RESTARTS   AGE
+   redis-59c9d56fc-6gcbc         2/2     Running   0          2m8s
+   demo-app-5845d6447b-v7npw     2/2     Running   0          2m8s
+   kuma-4j6wr-58998b5576-25wl6   1/1     Running   0          30s
+   ```
+   
+   ```sh
+   kubectl get svc -n kuma-demo
+   ```
+   
+   ```
+   NAME         TYPE           CLUSTER-IP      EXTERNAL-IP   PORT(S)          AGE
+   redis        ClusterIP      10.43.223.223   <none>        6379/TCP         3m27s
+   demo-app     ClusterIP      10.43.216.203   <none>        5000/TCP         3m27s
+   kuma-pfh4s   LoadBalancer   10.43.122.93    172.20.0.3    8080:30627/TCP   87s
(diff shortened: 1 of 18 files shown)


app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
```diff
@@ -262,19 +271,47 @@ spec:
...
+### Multi-zone Deployments
 
+{% if_version lte:2.6.x %}
 Gateway API isn't supported with multi-zone deployments, use {{site.mesh_product_name}}'s [`MeshGateways`](/docs/{{ page.version }}/using-mesh/managing-ingress-traffic/builtin-listeners)/[`MeshHTTPRoute`](/docs/{{ page.version }}/policies/meshhttproute)/
 [`MeshTCPRoute`](/docs/{{ page.version }}/policies/meshtcproute) instead.
+{% endif_version %}
+{% if_version gte:2.7.x %}
+Gateway API supports multi-zone deployments with some limitations.
+
+#### Important Considerations:
```
cloudjumpercat, line 296:
```suggestion
The Gateway API supports multi-zone deployments, but with some limitations:
```

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
```diff
@@ -262,19 +271,47 @@ spec:
...
+
+- Gateway API resources like `Gateway`, `ReferenceGrant`, and `HTTPRoute` must be created in non-global zones. They are specific to the zone they reside in.
+- Only services deployed within the same kubernetes cluster as the `HTTPRoute` can be referenced as the `backendRef`.
```
cloudjumpercat, line 299:
```suggestion
- Only services deployed within the same Kubernetes cluster, such as the `HTTPRoute`, can be referenced as the `backendRef`.
```

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
```diff
@@ -262,19 +271,47 @@ spec:
...
+{% capture backendref-limitation %}
+{% tip %}
+This is temporary limitation, which will be lifted when [work on allowing targeting {{site.mesh_product_name}}'s `MeshService`s in `backendRef`](https://github.com/kumahq/kuma/issues/9894) will be completed.
+{% endtip %}
+{% endcapture %}
+{{ backendref-limitation | indent }}
+
+   **Scenario:**
+   
+   Imagine you have two zones (`zone-1` in one kubernetes cluster and `zone-2` separate one) with a service named `backend` deployed in each zone and service `db` deployed only in `zone-1`. Here's what happens:
```
cloudjumpercat, line 310:
```suggestion
Here's an example scenario that describes how you could configure multi-zone deployments with the Gateway API. In this example, you have the following resources:
* Two zones:
  * `zone-1` is in a Kubernetes cluster
  * `zone-2` is in a separate Kubernetes cluster
*  Two services:
  * A service named `backend` deployed in each zone 
  * A service named `db` deployed only in `zone-1`

If you deploy multi-zone with Gateway API, the following will occur:
```

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
```diff
@@ -262,19 +271,47 @@ spec:
...
+   
+   - If you create an `HTTPRoute` with a `backendRef` targeting the backend service in `zone-1`, it will only route traffic to the `backend` service within `zone-1`.
```
cloudjumpercat, line 312:
```suggestion
   - If you create an `HTTPRoute` with a `backendRef` targeting the backend service in `zone-1`, it will only route traffic to the `backend` service in `zone-1`.
```

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
```diff
@@ -262,19 +271,47 @@ spec:
...
+   - Similarly, if you create an `HTTPRoute` in zone-2 with a `backendRef` pointing to the `db` service, it will result in `HTTPRoute` with a `ResolvedRefs` status condition of `BackendNotFound` because service `db` is not present in `zone-1`.
```
cloudjumpercat, line 313:
```suggestion
   - Similarly, if you create an `HTTPRoute` in `zone-2` with a `backendRef` pointing to the `db` service, it will result in a `HTTPRoute` with a `ResolvedRefs` status condition of `BackendNotFound` because service `db` is not present in `zone-1`.
```

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
```diff
@@ -262,19 +271,47 @@ spec:
...
+{% endif_version %}
```
cloudjumpercat, line 314: A couple other edits you could make:

- I think adding a Mermaid diagram that visually shows how the deployment scenario would look might help users understand it better.
- The other sections had an example yaml configuration, I think having something similar that users could look at and copy/paste to make their own would be helpful.
- I might be wrong, but it sounds like the example scenario you described is an example where the deployment isn't working correctly/as expected. Could we add another example, especially with an example yaml of a recommended configuration where things would work correctly/as expected so that users can see how they should be configuring this? TL;DR, it might be good to explain what to do in addition to what not to do.

cloudjumpercat (commented):
@bartsmykla Overall, the content in the section you added looks good! I just left some suggested formatting changes as well as a few other things you could edit/add if you wanted. Let me know if you have any questions about any of the feedback!

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
```diff
@@ -262,19 +271,51 @@ spec:
...
+{% if_version gte:2.7.x %}
+The Gateway API supports multi-zone deployments, but with some limitations:
+
+- Gateway API resources like `Gateway`, `ReferenceGrant`, and `HTTPRoute` must be created in non-global zones. They are specific to the zone they reside in.
+- Only services deployed within the same Kubernetes cluster, such as the `HTTPRoute`, can be referenced as the `backendRef`.
+
+{% capture backendref-limitation %}
+{% tip %}
+This is temporary limitation, which will be lifted when [work on allowing targeting {{site.mesh_product_name}}'s `MeshService`s in `backendRef`](https://github.com/kumahq/kuma/issues/9894) will be completed.
```
slonka, line 301: nitpick:

```suggestion
This is a temporary limitation, which will be lifted when [work on allowing targeting {{site.mesh_product_name}}'s `MeshService`s in `backendRef`](https://github.com/kumahq/kuma/issues/9894) is completed.
```

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
```diff
@@ -9,20 +9,26 @@ for configuring {% if_version gte:2.6.x inline:true %}[built-in gateway](/docs/{
...
-
+{% endwarning %}
+{% warning %}
 Gateway API [`Gateways`](https://gateway-api.sigs.k8s.io/api-types/gateway/) aren't supported in multi-zone. To use the builtin Gateway, you need to use the [`MeshGateway` resources](/docs/{{ page.version }}/using-mesh/managing-ingress-traffic/builtin-listeners).
 {% endwarning %}
 
 1. Install the Gateway API CRDs.
 
-   The Gateway API CRDs aren't available in Kubernetes by default yet. You must first [install the _experimental release_ that includes those CRDs as well as the admission webhook](https://gateway-api.sigs.k8s.io/guides/#install-experimental-channel).
+   Kubernetes doesn't include Gateway API functionality by default. To use it, you'll need to install the CRDs first. Instructions: [Installing Gateway API Standard Channel](https://gateway-api.sigs.k8s.io/guides/#install-standard-channel).
```
slonka, line 22: nitpick: to not duplicate this in line 30 everything above this could be inside `{% if_version lte:2.6.x %}` and start another block with `{% if_version lte:2.6.x %}` after this line.

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
```diff
@@ -262,19 +271,51 @@ spec:
...
+{% tip %}
+This is temporary limitation, which will be lifted when [work on allowing targeting {{site.mesh_product_name}}'s `MeshService`s in `backendRef`](https://github.com/kumahq/kuma/issues/9894) will be completed.
+{% endtip %}
+{% endcapture %}
+{{ backendref-limitation | indent }}
+
+Here's an example scenario that describes how you could configure multi-zone deployments with the Gateway API. In this example, you have the following resources:
+* Two zones:
+  * `zone-1` is in a Kubernetes cluster
+  * `zone-2` is in a separate Kubernetes cluster
```
slonka, line 309: so both of them are in separate cluster right? (a bit confusing that only in `zone-2` you mention that it's in a separate cluster) so maybe:

```suggestion
* Two zones (`zone-1` and `zone-2`) in separate Kubernetes clusters
```

?

slonka (commented):
A couple of things, also might be good to add the things that Diana mentioned. Good work 👍

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
bartsmykla: Then we would end with a numbered list containing only one element which I want to avoid

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
```diff
@@ -262,19 +271,49 @@ spec:
...
+{% capture backendref-limitation %}
+{% tip %}
+This is a temporary limitation, which will be lifted when [work on allowing targeting {{site.mesh_product_name}}'s `MeshService`s in `backendRef`](https://github.com/kumahq/kuma/issues/9894) is completed.
+{% endtip %}
+{% endcapture %}
+{{ backendref-limitation | indent }}
+
+Here's an example scenario that describes how you could configure multi-zone deployments with the Gateway API. In this example, you have the following resources:
+* Two zones (`zone-1` and `zone-2`) in separate Kubernetes clusters
+*  Two services:
```
bartsmykla, line 308:
```suggestion
* Two services:
```

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
lahabana, line 22:
```suggestion
   Kubernetes doesn't include Gateway CRDs by default. To use it, you'll need to install the CRDs first. Instructions: [Installing Gateway API Standard Channel](https://gateway-api.sigs.k8s.io/guides/#install-standard-channel).
```

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
```diff
@@ -9,20 +9,26 @@ for configuring {% if_version gte:2.6.x inline:true %}[built-in gateway](/docs/{
...
-   The Gateway API CRDs aren't available in Kubernetes by default yet. You must first [install the _experimental release_ that includes those CRDs as well as the admission webhook](https://gateway-api.sigs.k8s.io/guides/#install-experimental-channel).
+   Kubernetes doesn't include Gateway API functionality by default. To use it, you'll need to install the CRDs first. Instructions: [Installing Gateway API Standard Channel](https://gateway-api.sigs.k8s.io/guides/#install-standard-channel).
 
 2. Enable Gateway API support.
 
    - With `kumactl`, use the `--experimental-gatewayapi` flag.
    - With Helm, use the `{{site.set_flag_values_prefix}}experimental.gatewayAPI=true` value.
+{% endif_version %}
+{% if_version gte:2.7.x %}
+Kubernetes doesn't include Gateway API functionality by default. To use it, you'll need to install the CRDs first. Instructions: [Installing Gateway API Standard Channel](https://gateway-api.sigs.k8s.io/guides/#install-standard-channel).
```
lahabana, line 30:
```suggestion
Kubernetes doesn't include Gateway API CRDs by default. To install them follow the [Installing Gateway API Standard Channel](https://gateway-api.sigs.k8s.io/guides/#install-standard-channel).
```

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
lahabana: +1 to mermaid diagram!

Also can we make this work by using a service with `ExternalName` https://kubernetes.io/docs/concepts/services-networking/service/#externalname ?

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
slonka: why? we don't have to have the numbered item in the if, right? anyway, just a nitpick

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
bartsmykla: > Also can we make this work by using a service with `ExternalName` https://kubernetes.io/docs/concepts/services-networking/service/#externalname ?

Unfortunately not:
> ExternalName services can refer to CNAME DNS records that may live outside of the cluster and as such are difficult to reason about in terms of conformance. They also may not be safe to forward to (see CVE-2021-25740 for more information). Implementations SHOULD NOT support ExternalName Services.
>
> ref. https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.BackendObjectReference

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
```diff
@@ -293,28 +293,123 @@ Gateway API isn't supported with multi-zone deployments, use {{site.mesh_product
...
+            end
+            route1--"❌"-->db
+            linkStyle 0 stroke:red,color:red,stroke-dasharray: 5 5;
+            listener-->route1
+        end
+    end
+{% endmermaid %}
+{% endcapture %}
+
+   - If in `k8s-cluster-1` you create an `HTTPRoute` with a `backendRef` targeting the `backend` service, it will only route traffic to the `backend` service in `k8s-cluster-1`.
```
slonka, line 405: Shouldn't it be

```suggestion
   - If you create an `HTTPRoute` with a `backendRef` targeting the `backend` service in `k8s-cluster-1`, it will only route traffic to the `backend` service in `k8s-cluster-1`.
```

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
```diff
@@ -293,28 +293,123 @@ Gateway API isn't supported with multi-zone deployments, use {{site.mesh_product
...
+     
+     {{ gapi_multizone_limitation_1 | indent }}
+
+   - Similarly, if in `k8s-cluster-1` you create an `HTTPRoute` with a `backendRef` pointing to the `db` service, it will result in a `HTTPRoute` with a `ResolvedRefs` status condition of `BackendNotFound` because service `db` is not present in `k8s-cluster-1`.
```
slonka, line 409:
```suggestion
   - Similarly, if you create an `HTTPRoute` with a `backendRef` pointing to the `db` service in `k8s-cluster-1`, it will result in a `HTTPRoute` with a `ResolvedRefs` status condition of `BackendNotFound` because service `db` is not present in `k8s-cluster-1`.
```

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
slonka: > I might be wrong, but it sounds like the example scenario you described is an example where the deployment isn't working correctly/as expected. Could we add another example, especially with an example yaml of a recommended configuration where things would work correctly/as expected so that users can see how they should be configuring this? 

I don't think this part was addressed yet, right @bartsmykla ?

slonka (commented):
Diagrams look good, added two suggestions for the text and one thing is not yet addressed (I think)

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
```diff
@@ -9,20 +9,26 @@ for configuring {% if_version gte:2.6.x inline:true %}[built-in gateway](/docs/{
...
-
+{% endwarning %}
+{% warning %}
 Gateway API [`Gateways`](https://gateway-api.sigs.k8s.io/api-types/gateway/) aren't supported in multi-zone. To use the builtin Gateway, you need to use the [`MeshGateway` resources](/docs/{{ page.version }}/using-mesh/managing-ingress-traffic/builtin-listeners).
 {% endwarning %}
 
 1. Install the Gateway API CRDs.
 
-   The Gateway API CRDs aren't available in Kubernetes by default yet. You must first [install the _experimental release_ that includes those CRDs as well as the admission webhook](https://gateway-api.sigs.k8s.io/guides/#install-experimental-channel).
+   Kubernetes doesn't include Gateway API CRDs by default. To install them follow the [Installing Gateway API Standard Channel](https://gateway-api.sigs.k8s.io/guides/#install-standard-channel).
```
michaelbeaumont, line 22:
```suggestion
   Kubernetes doesn't include Gateway API CRDs, [install them from the standard channel CRD bundle.](https://gateway-api.sigs.k8s.io/guides/#install-standard-channel).
```

and below of course.

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
```diff
@@ -262,19 +271,146 @@ spec:
...
+### Multi-zone Deployments
 
+{% if_version lte:2.6.x %}
 Gateway API isn't supported with multi-zone deployments, use {{site.mesh_product_name}}'s [`MeshGateways`](/docs/{{ page.version }}/using-mesh/managing-ingress-traffic/builtin-listeners)/[`MeshHTTPRoute`](/docs/{{ page.version }}/policies/meshhttproute)/
 [`MeshTCPRoute`](/docs/{{ page.version }}/policies/meshtcproute) instead.
+{% endif_version %}
+{% if_version gte:2.7.x %}
+The Gateway API supports multi-zone deployments, but with some limitations:
+
+- Gateway API resources like `Gateway`, `ReferenceGrant`, and `HTTPRoute` must be created in non-global zones. They are kubernetes resources and therefore are not known to {{ site.mesh_product_name }}.
```
michaelbeaumont, line 296:
```suggestion
- Gateway API resources like `Gateway`, `ReferenceGrant`, and `HTTPRoute` must be created in non-global zones}. They also don't appear directly in the {{ site.mesh_product_name }} GUI, only the {{ site.mesh_product_name }} resources they manage do.
```

Not sure what they are "not known to Kuma" means?

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
```diff
@@ -262,19 +271,146 @@ spec:
...
+- Only services deployed within the same Kubernetes cluster, such as the `HTTPRoute`, can be referenced as the `backendRef`.
```
michaelbeaumont, line 297:
```suggestion
- Only services deployed within the same Kubernetes cluster, such as the `HTTPRoute`, can be referenced via `backendRef`.
```

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
```diff
@@ -287,32 +423,34 @@ Gateway API isn't supported with multi-zone deployments, use {{site.mesh_product
 GAMMA is a Gateway API subproject focused on mesh implementations of Gateway API
 and extending the Gateway API resources to mesh use cases.
 
+{% if_version lte:2.6.x %}
 {% warning %}
-GAMMA itself as well as GAMMA in Kuma are **experimental**!
+GAMMA in {{site.mesh_product_name}} is **experimental**!
```
michaelbeaumont, line 428:
```suggestion
GAMMA in {{site.mesh_product_name}} is **experimental**!
```

Is it? Maybe beta?

bartsmykla: It's wrapped in `{% if_version lte:2.6.x %}` block, so <= 2.6.x it's `experimental` >= 2.7.0 it's not

michaelbeaumont: Ah right, why change the message at all? Or at least maybe make it clear that at the time, GAMMA was also not yet GA?

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
bartsmykla: It works as expected, but it might be confusing, as in Gateway API scenario we are operating on kubernetes resources, not ours.

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
```diff
@@ -9,20 +9,26 @@ for configuring {% if_version gte:2.6.x inline:true %}[built-in gateway](/docs/{
...
-   The Gateway API CRDs aren't available in Kubernetes by default yet. You must first [install the _experimental release_ that includes those CRDs as well as the admission webhook](https://gateway-api.sigs.k8s.io/guides/#install-experimental-channel).
+   Kubernetes doesn't include Gateway API CRDs by default. To install them follow the [Installing Gateway API Standard Channel](https://gateway-api.sigs.k8s.io/guides/#install-standard-channel).
 
 2. Enable Gateway API support.
 
    - With `kumactl`, use the `--experimental-gatewayapi` flag.
    - With Helm, use the `{{site.set_flag_values_prefix}}experimental.gatewayAPI=true` value.
+{% endif_version %}
+{% if_version gte:2.7.x %}
+Kubernetes doesn't include Gateway API CRDs by default. To install them follow the [Installing Gateway API Standard Channel](https://gateway-api.sigs.k8s.io/guides/#install-standard-channel).
```
bartsmykla, line 30:
```suggestion
Kubernetes doesn't include Gateway API CRDs, [install them from the standard channel CRD bundle](https://gateway-api.sigs.k8s.io/guides/#install-standard-channel).
```

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
bartsmykla: I actually removed this sentence as it didn't make sense to me anymore.

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
michaelbeaumont: I guess in this case I would prefer that when looking at the docs for an older version I would see that the implementation, in that version, is for an experimental spec.

michaelbeaumont: But not a big deal

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
bartsmykla: I changed the wording a little bit, should be better now

michaelbeaumont (approved):

slonka (approved):

PR #1743, opened 2024-04-15
fix(docs): ensure getting started links have a return between them

johncowen:
Looks like there is some strange markdown behaviour with `if_version`.

This PR just amends one problem, but doesn't solve whatever the issue is. Happy to close if someone has a better solution.

### Before

Screenshot 2024-04-15 at 16 19 50

### After

Screenshot 2024-04-15 at 16 19 36

### DIFF
--- a/app/_src/index.md
+++ b/app/_src/index.md
@@ -28,6 +28,7 @@ The core maintainer of Kuma is **Kong**, the maker of the popular open-source Ko
 {% if_version gte:2.2.x %}
 [Read about service mesh](/docs/{{ page.version }}/introduction/about-service-meshes)
 
+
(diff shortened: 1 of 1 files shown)


slonka:
Still renders the wrong way in 2.1:

image

https://deploy-preview-1743--kuma.netlify.app/docs/2.1.x/

slonka (approved):

slonka:
FYI @fabianrbz xrel: https://github.com/Kong/jekyll-generator-single-source/pull/16

slonka:
@johncowen - added a newline to your PR so it renders correctly in older versions

PR #1759, opened 2024-04-22
fix(docs): fix some typos

Icarus9913:
Fix some document typos

Did you sign your commit? [Instructions](https://github.com/kumahq/.github/blob/main/CONTRIBUTING.md#sign-your-commits)- Yes

Have you read [Contributing guidelines](https://github.com/kumahq/.github/blob/main/CONTRIBUTING.md)?- Yes

### DIFF
--- a/app/_src/introduction/architecture.md
+++ b/app/_src/introduction/architecture.md
@@ -135,7 +135,7 @@ Kubernetes `Jobs` are a good example of this.
 
 In this case, the {{site.mesh_product_name}} control plane generates a `kuma.io/service` tag with the format `<name>_<namespace>_svc`, where `<name>` and`<namespace>` come from the `Pod` resource itself.
 
-The `Pods` created by the following example `Deployment` have the tag `kuma.io/service: example-client_kuma-example_svc`:
+The `Pods` created by the following example `Deployment` have the tag `kuma.io/service: echo-client_default_svc`:
 
 ```yaml
 apiVersion: apps/v1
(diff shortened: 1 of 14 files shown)


app/_src/production/init-containers.md
```diff
@@ -6,9 +6,9 @@ Due to the way that {{site.mesh_product_name}} implements transparent proxying,
 
 ### Network calls to outside of the mesh
 
-The common pitfall is the idea that it's possible to order init containers so that the mesh init container is run after other init containers.
+The common pitfall is the idea that it's possible to order init containers so that the mesh init container is running after other init containers.
```
slonka, line 9: I think the old version is correct but I'm not a native speaker.

app/_src/production/init-containers.md
```diff
@@ -6,9 +6,9 @@ Due to the way that {{site.mesh_product_name}} implements transparent proxying,
...
 However, when injecting these init containers into a Pod via webhooks, such as the Vault init container, there is no assurance of the order.
-The ordering of init containers also doesn't provide a solution when the CNI is used, as traffic redirection to the sidecar occurs even before any init container runs
+The ordering of init containers also doesn't provide a solution when the Kuma CNI is used, as traffic redirection to the sidecar occurs even before any init container runs.
```
slonka, line 11: If you really want to put the product name there you should you `{{ site.mesh_product_name }}` so when it's included in [kong-mesh docs](https://docs.konghq.com/mesh/latest/) it's rendered correctly:

```suggestion
The ordering of init containers also doesn't provide a solution when the {{ site.mesh_product_name }} CNI is used, as traffic redirection to the sidecar occurs even before any init container runs.
```

Icarus9913: Got it, thank you

app/_src/production/init-containers.md
Icarus9913: Yeah, the old version is correct. It's a passive pattern.

lahabana (approved):

app/_src/production/init-containers.md
lahabana: `is running` is easier to understand IMO

johncowen: How about just:

> The common pitfall is the idea that it's possible to order init containers so that the mesh init container runs after other init containers.

I asked the copywriter sat next to me 😆 and that was the suggestion

PR #1760, opened 2024-04-22
chore(docs): optimize the local running docs

Icarus9913:
The command `make install` is not obvious in the `CONTRIBUTING.md`, and I just make it equal with other commands

Did you sign your commit? [Instructions](https://github.com/kumahq/.github/blob/main/CONTRIBUTING.md#sign-your-commits) - yes

Have you read [Contributing guidelines](https://github.com/kumahq/.github/blob/main/CONTRIBUTING.md)? - yes

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,7 +61,11 @@ In most cases, you can add this signoff to your commit automatically with the `-
 
 ### Running the site locally
 
-Clone the repository and run `make install`, make sure you have the right ruby version installed (`cat .ruby-version`).
+Clone the repository and run:
+```
+make install
+```
+make sure you have the right ruby version installed (`cat .ruby-version`).
 Jekyll is a static-site generator, so first we need to build the site and compile the assets:
 ```
 make build


lahabana (approved):

.gitignore
```diff
@@ -3,6 +3,7 @@ node_modules
 .idea
 docs/.vuepress/dist
 yarn-error.log
+yarn.lock
```
johncowen, line 6: Bit of a drive by review but glanced and saw this, should we be ignoring the yarn.lock file?

Icarus9913: I don't know...
I just run `make install && make build` and found this file. I mean it's a generated file from compile, maybe we could  ignore it?

johncowen: Usually a `yarn.lock` is commited to pin dependencies, but sounds like here that has never been done. I don't know enough about these build/install scripts to know if there is a reason for this, so same as you! 😅 😆 🙌 

Umm, wow it does `npm install` then `yarn install`, ideally we should probably take look at this. Lemme just check with folks but at the least I think we shouldn't ignore the lock file just yet.

johncowen: Ok yeah so we _are_ pinning our dependencies here (if I make a change to the lock file git tells me).

We definitely shouldn't gitignore it.

Icarus9913: OK, got it.

kleinfreund: > Umm, wow it does `npm install` then `yarn install`

😬

kleinfreund: Yeah, this should be changed:

```diff
- 	npm install -g netlify-cli@16.5.1
+ 	yarn global add netlify-cli@16.5.1 
```

jijiechen (approved):

johncowen (approved):

Icarus9913:
@johncowen @kleinfreund 
Hey, could you please help me to merge it? I and Jay have no rights to merge this PR.

PR #1772, opened 2024-04-24
chore: install netlify-cli in project instead of globally

kleinfreund:
Install the netlify-cli package in the project instead of globally.

Note: the packages in package.json are _very_ outdated. It doesn’t look like we’re using Dependabot here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,6 @@ endif
 
 # Installs npm packages and gems.
 install: ruby-version-check
-	npm install -g netlify-cli@16.5.1
 	yarn install
 	bundle install
 
@@ -27,7 +26,7 @@ build: ruby-version-check
 	bundle exec jekyll build --config jekyll.yml --profile
 
 serve:
-	netlify dev
+	yarn run dev
 
 # Cleans up all temp files in the build.
 # Run `make clean` locally whenever you're updating dependencies, or to help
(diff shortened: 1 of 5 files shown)


johncowen (approved):
Thanks!

johncowen:
Oh actually, could you have a look and see where else netlify is mentioned in the project? I think there are a couple of other places that should now use `serve` or `yarn dev`

kleinfreund:
> Oh actually, could you have a look and see where else netlify is mentioned in the project? I think there are a couple of other places that should now use `make serve` or `yarn dev`

I did and I could only find this one instance. Have I missed any that you know of?

kleinfreund:
@johncowen just pinging you in case you missed this. If you don’t know of any more instances that I haven’t found, I’ll merge this in a while.

kleinfreund:
> Does the one in the `Procfile` need changing?

Ah, interesting. I didn’t find this because I’ve already had the netlify line removed for a long time locally (I needed that for some reason to workaround some issue) and so didn’t consider it at all. I’ve updated it now.

johncowen:
Ok cool, can you ping someone else to check this real quick, the JS here is fine by me, but I'm not totally sure about the Procfile and what its used for etc (especially seeing as you didn't need it locally)

kleinfreund:
> Ok cool, can you ping someone else to check this real quick, the JS here is fine by me, but I'm not totally sure about the Procfile and what its used for etc (especially seeing as you didn't need it locally)

@fabianrbz Could you have a look at this, please?

fabianrbz (approved):

johncowen:
thanks @fabianrbz 👍

PR #1774, opened 2024-04-25
fix: minor suggestions for modifications to "how mesh works"

RoRoJ:
This was a great doc for understanding more about the Service Mesh, but I noticed a few minor grammar/formulation points to suggest modifications to.


lahabana (approved):

lahabana:
Thanks a lot for the contrib!

PR #1782, opened 2024-05-07
docs(transparentproxy): add info about adjusting exclude inbound/outbound ports

slonka:
this came as a question on slack


app/_src/production/dp-config/transparent-proxying.md
```diff
@@ -60,7 +60,7 @@ The host that will run the `kuma-dp` process in transparent proxying mode needs
    useradd -u 5678 -U kuma-dp
    ```
 
-3. Redirect all the relevant inbound, outbound and DNS traffic to the {{site.mesh_product_name}} data plane proxy.
+3. Redirect all the relevant inbound, outbound and DNS traffic to the {{site.mesh_product_name}} data plane proxy (if you're running any other services on that machine you need to adjust the comma separated lists of `--exclude-inbound-ports` and `--exclude-outbound-ports` accordingly).
```
lahabana, line 63: Isn't the right config something with uid too?

slonka: would you like to exclude it from redirects completely? like with `--kuma-dp-uid`? we'd need another switch for that like `--redirect-disabled-uids`?

slonka: @lahabana WDYT?

Automaat (approved):

app/_src/production/dp-config/transparent-proxying.md
slonka: let's discuss it in an issue https://github.com/kumahq/kuma/issues/10672

PR #1783, opened 2024-05-07
docs: update Envoy requirements

### LINKED ISSUE #1776: The description of Envoy requirement is confusing.
### What happened?

In https://kuma.io/docs/2.7.x/introduction/kuma-requirements/#envoy,

> Versions of envoy supported are: ~1.25.0 which means >=1.25.0 and <1.25.0.
> You can find the current version in the [version.sh](https://github.com/kumahq/kuma/blob/master/tools/releases/version.sh#L11) file.

For the first sentence, if the supported Envoy is both >=1.25.0 and <1.25.0, it means we support every version of Envoy. Maybe this sentence misses a patch version?

And what's the difference between the supported version and the currently used version? I find that the current used version is 1.29.4, which is much newer than the supported one.

spacewander:
Fix #1776
Signed-off-by: spacewander <spacewanderlzx@gmail.com>


lahabana (approved):
Thx for the contrib!

PR #1788, opened 2024-05-13
chore(ci): remove installer test for hardcoded preview version

Automaat:
Remove CI step with hardcoded preview version, as we are basically testing this in other steps

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -136,11 +136,11 @@ jobs:
           VERBOSE: ${{ runner.debug }}
           VERSION: preview
         run: |
-          export VERSION=$(sh app/installer.sh --print-version 2> /dev/null)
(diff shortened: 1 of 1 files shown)


michaelbeaumont (changes requested):

michaelbeaumont (approved):

.github/workflows/ci.yml
```diff
@@ -134,13 +134,13 @@ jobs:
       - name: Run installer.sh with preview version
         env:
           VERBOSE: ${{ runner.debug }}
-          VERSION: 0.0.0-preview.v76e03eb68
+          VERSION: preview
```
slonka, line 137: this does not take any effect since you're overriding this in line 139, we should have a test with `preview` as a version as well

PR #1862, opened 2024-07-29
fix(zone-ingress): use same port in zoneIngress example

### LINKED ISSUE #1861: Issue with port config on zone-ingress page
### What happened?

https://kuma.io/docs/2.8.x/production/cp-deployment/zone-ingress/#zone-ingress

The port config in the `[Universal]` tab `10000` and `10001` are problematic.

Screenshot 2024-07-26 at 12 39 53

In quick setups like demos we usually don't add a LB in front of the ingress and just run a single zoneIngress. Even when using a LB it would be weird to use a different port

lahabana:
fix #1861

### DIFF
--- a/app/_src/production/cp-deployment/zone-ingress.md
+++ b/app/_src/production/cp-deployment/zone-ingress.md
@@ -63,7 +63,7 @@ networking:
   address: 192.168.0.1
   port: 10001
   advertisedAddress: 10.0.0.1
-  advertisedPort: 10000
+  advertisedPort: 10001
 ```
 
 Apply the ingress configuration, passing the IP address of the control plane and your instance should start.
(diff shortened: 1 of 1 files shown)


michaelbeaumont (commented):
What exactly is the problem here? None of the issues or notes mention what's wrong.

lahabana:
@michaelbeaumont added info in the issue but also comments in the file to explain what the `advertised*` are for

michaelbeaumont (approved):
So it's that it's easier to copy paste? Why is the address different then? But :shrug:

lahabana:
> So it's that it's easier to copy paste? Why is the address different then? But 🤷
I'm not sure though for sure in practice the address will be different though port should likely be the same right?

Issue #1890, opened 2024-09-09
MeshTLS docs


slonka:
### Description

Continuation from https://github.com/kumahq/kuma/issues/11134

PR #1893, opened 2024-09-10
feat(meshtls): initial docs

slonka:
rendered: https://deploy-preview-1893--kuma.netlify.app/docs/dev/policies/meshtls/

closes https://github.com/kumahq/kuma-website/issues/1890

### DIFF
--- a/app/_src/policies/meshtls.md
+++ b/app/_src/policies/meshtls.md
@@ -36,6 +36,13 @@ The following describes the default configuration settings of the `MeshTLS` poli
 - **`tlsCiphers`**: Defines TLS ciphers to be used by **both client and server**. Allowed values: `ECDHE-ECDSA-AES128-GCM-SHA256`, `ECDHE-ECDSA-AES256-GCM-SHA384`, `ECDHE-ECDSA-CHACHA20-POLY1305`, `ECDHE-RSA-AES128-GCM-SHA256`, `ECDHE-RSA-AES256-GCM-SHA384`, `ECDHE-RSA-CHACHA20-POLY1305`.
 - **`mode`**: Defines the mTLS mode - `Permissive` mode encrypts outbound connections the same way as `Strict` mode, but inbound connections on the server-side accept both TLS and plaintext. Allowed values: `Strict`, `Permissive`.
 
+{% tip %}
+Setting the TLS version and ciphers on both the client and server makes it harder to misconfigure.
+If you want to try out a specific version/cipher combination, we recommend creating a [temporary mesh](/docs/{{ page.version }}/production/mesh/#usage), deploying two applications within it, and testing whether communication is working.
+If you have a use case for configuring a different set of allowed versions/ciphers on different workloads, we'd love to hear about it.
+In that case, please open an [issue](https://github.com/kumahq/kuma/issues).
+{% tip %}
+
 ## Examples
 
 ### Set specific TLS version and ciphers
--- a/app/_src/policies/meshtls.md
+++ b/app/_src/policies/meshtls.md
@@ -82,4 +82,4 @@ spec:
 
 ## All policy options
 
-{% json_schema MeshTLS %}
+{% json_schema MeshTLSes %}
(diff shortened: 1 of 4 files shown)


app/_src/policies/meshtls.md
```diff
@@ -0,0 +1,85 @@
+---
+title: MeshTLS
+---
```
lahabana, line 3: Should we add a guide along with this named something like: "Progressively rolling in strict MTLS" and then we could link it from this page?

Something that's not clear is how clients get influenced by the setting of MeshTLS. We had a discussion about this earlier this week, we should really document how this works.

app/_src/policies/meshtls.md
```diff
@@ -0,0 +1,85 @@
...
+
+{% warning %}
+This policy uses new policy matching algorithm.
+{% endwarning %}
+
+This policy enables {{site.mesh_product_name}} to configure TLS mode, ciphers and version. 
+Backends and default mode values are taken from [the Mesh object](/docs/{{ page.version }}/policies/mutual-tls/).
```
lahabana, line 10: Should you talk about this policy in the page above? Seems like it no?

app/_src/policies/meshtls.md
slonka: created a separate issue for this https://github.com/kumahq/kuma-website/issues/1897

slonka: > Something that's not clear is how clients get influenced by the setting of MeshTLS. We had a discussion about this earlier this week, we should really document how this works.

In line https://github.com/kumahq/kuma-website/pull/1893/files#diff-9149b3f0d9714aeffecfdede60cbc8469e9593e7acee44002e0d0aa61444b201R35 I'm mentioning that we configure both server and client but I'll make it more explicit.

app/_src/policies/meshtls.md
slonka: added a link in `policies/mutual-tls`

app/_src/policies/meshtls.md
slonka: added something like this: https://github.com/kumahq/kuma-website/pull/1893/files#diff-9149b3f0d9714aeffecfdede60cbc8469e9593e7acee44002e0d0aa61444b201R40

app/_src/policies/mutual-tls.md
```diff
@@ -2,6 +2,12 @@
 title: Mutual TLS
 ---
 
+{% if_version gte:2.9.x %}
+{% tip %}
+If you want to configure version, ciphers or per service permissive / strict mode check out [`MeshTLS`](/docs/{{ page.version }}/policies/mesh)
```
lahabana:
```suggestion
If you want to configure version, ciphers or per service permissive / strict mode check out [`MeshTLS`](/docs/{{ page.version }}/policies/meshtls)
```

app/_src/policies/meshtls.md
```diff
@@ -0,0 +1,92 @@
+---
+title: MeshTLS
+---
+
+{% warning %}
+This policy uses new policy matching algorithm.
```
lobkovilya, line 6: Do we really need this warning? We're using it in other policies so people don't combine them with legacy twin policies (i.e. Retry and MeshRetry). But MeshTLS is kind of unique so I think we don't need it.

app/_src/policies/meshtls.md
```diff
@@ -0,0 +1,92 @@
...
+{% policy_yaml example2 %}
+```yaml
+type: MeshTLS
+name: strict-mode
+mesh: default
+spec:
+  targetRef:
+    kind: MeshSubset
+    tags:
+      kuma.io/service: redis_kuma-demo_svc_6379
```
lobkovilya, line 81: Should we use `app: redis` tags? I think it's more future-proof than `kuma.io/service` tag

app/_src/policies/meshtls.md
```diff
@@ -0,0 +1,92 @@
+---
+title: MeshTLS
+---
+
+{% warning %}
+This policy uses new policy matching algorithm.
+{% endwarning %}
+
```
slonka, line 8:
```suggestion
```

lobkovilya (approved):

Issue #1897, opened 2024-09-16
MeshTLS guide


slonka:
### Description

create a guide with something like: "Progressively rolling in strict MTLS" and then we could link it to docs page

PR #1906, opened 2024-09-23
feat(meshtls): initial guide

slonka:
closes https://github.com/kumahq/kuma-website/issues/1897

### DIFF
--- a/app/_src/guides/progressively-rolling-in-strict-mtls.md
+++ b/app/_src/guides/progressively-rolling-in-strict-mtls.md
@@ -1,0 +1,85 @@
+---
+title: Progressively rolling in strict mTLS
+---
+
+The new MeshTLS policy allows you to gradually migrate services to mutual TLS without dropping a packet.
+
+## Prerequisites
+- Completed [quickstart](/docs/{{ page.version }}/quickstart/kubernetes-demo/) to set up a zone control plane with demo application
+
+## Gradually bring another service into the mesh
+
+### Start a second demo app
+
+First we start a new demo setup in another namespace called `kuma-demo-migration`.
+The below command installs the demo one more time in another namespace:
+
+```bash
+curl -s https://raw.githubusercontent.com/kumahq/kuma-counter-demo/master/demo.yaml | 
+  sed "s#kuma.io/sidecar-injection: enabled#kuma.io/sidecar-injection: disabled#" |
+  sed "s#name: kuma-demo#name: kuma-demo-migration#" |
+  sed "s#redis.kuma-demo.svc.cluster.local#redis.kuma-demo-migration.svc.cluster.local#" |
+  sed "s#namespace: kuma-demo#namespace: kuma-demo-migration#" | kubectl apply -f -
+```
+
+### Enable permissive mode on redis
+
+We begin with preparing redis to start in [permissive](/docs/{{ page.version }}/policies/meshtls/#configuration) mode when deployed inside the mesh.
+To enable permissive mode we define this `MeshTLS` policy:
+
+```yaml
+apiVersion: kuma.io/v1alpha1
+kind: MeshTLS
+metadata:
+  name: redis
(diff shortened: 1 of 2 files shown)


app/_src/guides/progressively-rolling-in-strict-mtls.md
```diff
@@ -0,0 +1,104 @@
+---
+title: Progressively rolling in strict mTLS
+---
+
+The new [MeshTLS](/docs/{{ page.version }}/policies/meshtls/) policy allows you to gradually migrate services to mutual TLS without dropping a packet.
```
lahabana, line 5:
```suggestion
The [MeshTLS](/docs/{{ page.version }}/policies/meshtls/) policy allows you to gradually migrate services to mutual TLS without dropping a packet.
```

Docs should be intemporal

slonka:
> It would be good to have a mermaid diagram to show what things look like at each step

added, do you think I should also somehow show when things are in `strict` vs `permissive` mode?

> It would be good at the end to have a section: "Things to remember when migrating to strict TLS" with things like: "before turning something to strict check that this stat is no longer incrementing"

added that but need to check that metric... @johncowen - did we write down the metrics we want to use in the GUI to show TLS-ed / non TLS-ed requests / connections?

also do you have any other final points that we should add?

slonka:
@lahabana - probably shouldn't ask the users to calculate things by hand... maybe we should keep `tls_inspector` stats for now with a `*` that it's not 100% sure and point people to GUI once the feature linked by JC makes sense, WDYT?

lahabana:
Providing at least clarity as to which metrics users should like like seems ok to me no? What's being calculated here?

The graphs look good but would it be helpful to color the arrows depending on TLS and non-TLS traffic?

app/_src/guides/progressively-rolling-in-strict-mtls.md
```diff
@@ -0,0 +1,185 @@
...
+  targetRef:
+    kind: MeshSubset
+    tags:
+      app: redis
+      k8s.kuma.io/namespace: kuma-demo-migration
+  from:
+  - targetRef:
+      kind: Mesh
+    default:
+      mode: Permissive
```
lahabana, line 81: I think in other guides we take stuff that can be directly copy pasted. Also I think the namespace should be `{{ site.namespace }}` e.g: https://github.com/kumahq/kuma-website/blob/master/app/_src/guides/gateway-builtin.md#define-a-route-using-meshhttproute

lahabana (approved):
1 nit but LGTM. I've ran the guide and it worked.

However, the reliance on these stats is hard to comprehend as I show in this zoom clip:
https://konghq.zoom.us/clips/share/enLGWQHCv3HzEKlFWkEGEeuXdOcyDYdGTHu518KqCwf3z1ZnpuntGwhv2UO2CzLJPwkAxynTL04Uc6Bx-Js7xtxp.Cu8iOtgCRRiLZmBS

Is there anything we can do better there? If not should we just say we're going to improve this and link to the issue?

app/_src/guides/progressively-rolling-in-strict-mtls.md
slonka: yes, definitely!

app/_src/guides/progressively-rolling-in-strict-mtls.md
```diff
@@ -0,0 +1,185 @@
...
+
+We begin with preparing redis to start in [permissive](/docs/{{ page.version }}/policies/meshtls/#configuration) mode when deployed inside the mesh.
+To enable permissive mode we define this `MeshTLS` policy:
+
+```yaml
+apiVersion: kuma.io/v1alpha1
+kind: MeshTLS
+metadata:
+  name: redis
+  namespace: kuma-system
```
slonka, line 68:
```suggestion
  namespace: {{ site.namespace }}
```

PR #1938, opened 2024-10-07
docs(networking): hostname generator improvements

jakubdyszkiewicz:
Add info about default hostname generators.
Point to HostnameGenerator from the general DNS section.
Improve policy yaml so it does not add `kuma.io/mesh` for non-mesh policies.


app/_src/networking/hostnamegenerator.md
```diff
@@ -2,32 +2,139 @@
 title: HostnameGenerator
 ---
 
+{% if_version gte:2.9.x %}
```
michaelbeaumont, line 5: sorry i meant the removal of the warning, the if_version should be
```suggestion
{% if_version lte:2.8.x %}
```

jakubdyszkiewicz: duh!

app/_src/networking/hostnamegenerator.md
```diff
@@ -2,32 +2,139 @@
...
+  selector:
+    meshService:
+      matchLabels:
+        kuma.io/origin: zone
+  template: "{% raw %}{{ .DisplayName }}.svc.mesh.local{% endraw %}"
+```
+
+### Local MeshExternalService
+
+The following policy is automatically created on a zone control plane.
```
michaelbeaumont, line 36: Just my opinion but maybe each should have an explanation, since they're essentially examples explaining the resource, like "add a hostname `{{ .DisplayName }}.extsvc.mesh.local` for every MeshExternalService in the local zone"

jakubdyszkiewicz: how about now?

PR #1942, opened 2024-10-08
fix(gui): update screenshots

lahabana:
Updated with newer screenshots and added some cross links


johncowen:
I saw the tiniest of things, not sure how important you think it is, but incase you didn't notice:

In this image https://github.com/kumahq/kuma-website/blob/49c1307992053e533e8c9d8c84e806f949ca8f6d/app/assets/images/docs/gui/overview.png you can see the Netlify footer thing.

I reckon we could just crop that out if we don't want that there?

johncowen (approved):

PR #1943, opened 2024-10-08
fix(install): rework install page

### LINKED ISSUE #1921: "Install kumactl" page should be easier to find
### What happened?

At this moment the page that aggregates all the installation info https://kuma.io/install/2.8.x/ apparently doesn't have link to https://kuma.io/docs/2.8.x/production/install-kumactl/#install-kumactl

lahabana:
Rework the install page to direct towards the guides or how to install kumactl.

This makes more sense and is a strong call to action than the previous install pages

Fix #1921


lukidzi (approved):
🎉

PR #1967, opened 2024-10-12
fix(guide): improve progressively rolling in strict mtls doc

Icarus9913:
improve Guide `Progressively rolling in strict mTLS` doc


lahabana (approved):
one nit but LG

app/_src/guides/progressively-rolling-in-strict-mtls.md
```diff
@@ -50,10 +50,11 @@ flowchart LR
     linkStyle 1 stroke:#555a5d, stroke-width:2px;
 {% endmermaid %}
 
-### Enable port forwarding for both second app
+### Enable port forwarding for both demo-app
 
 ```bash
```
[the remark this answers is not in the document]
Icarus9913: @slonka Do you have better title here?

app/_src/guides/progressively-rolling-in-strict-mtls.md
```diff
@@ -175,13 +176,15 @@ flowchart LR
...
 Finally, to set strict mode you can either edit the policy or remove it (the default is taken from `Mesh` object which is `STRICT`).
 
-```bash
-kubectl delete meshtlses.kuma.io -n kuma-system redis
-```
+{% tip %}
+**Things to remember when migrating to strict TLS**
 
-## Things to remember when migrating to strict TLS
+Before changing a workload to `Strict` mode check that `tls_inspector.tls_not_found` stat is no longer incrementing.
```
slonka, line 182: we no longer use this stat to indicate this, we're using https://github.com/kumahq/kuma-website/pull/1967/files#diff-2f2247fada7fc4374636439b1747188dfc75fb57ae927de5bd76780696012f4dR101 so it should be something like the difference between `inbound_POD_IP_6379.rbac.allowed: 809` and `cluster.localhost_6379.upstream_cx_total` from up the point of the migration should  be 0. Not sure how to clearly indicate that.

slonka (changes requested):

app/_src/guides/progressively-rolling-in-strict-mtls.md
michaelbeaumont: We also sometimes say metric and sometimes stat. Which I think is probably pretty confusing, it should be always pointed out how to get this value. Since stats are not used in almost all mesh operations

app/_src/guides/progressively-rolling-in-strict-mtls.md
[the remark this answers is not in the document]
slonka: > You might want to say to do these in different terminal windows

yes

> Do you have better title here?
I think the title is fine, we could go with

> Enable port forwarding for both demo-apps

app/_src/guides/progressively-rolling-in-strict-mtls.md
```diff
@@ -5,7 +5,7 @@ title: Progressively rolling in strict mTLS
 The [MeshTLS](/docs/{{ page.version }}/policies/meshtls/) policy allows you to gradually migrate services to mutual TLS without dropping a packet.
 
 ## Prerequisites
-- Completed [quickstart](/docs/{{ page.version }}/quickstart/kubernetes-demo/) to set up a zone control plane with demo application
+- Completed [quickstart](/docs/{{ page.version }}/quickstart/kubernetes-demo/) to set up a zone control plane with demo application and make sure you enabled mTLS from the quickstart.
```
Automaat, line 8: Wouldn't it be better to just add the step with configuring mTLS in this guide? I do this in my namespace scoped policies and I think that we need something specific we should just reconfigure it in guide to make sure users have right config

app/_src/guides/progressively-rolling-in-strict-mtls.md
```diff
@@ -94,7 +95,7 @@ kubectl patch deployment redis -n kuma-demo-migration \
 ```
 
 After this redis will be receiving plaintext traffic from non-meshed client.
-You can go to {{site.mesh_product_name}} GUI (port 5681) and you should see this metric increment on `redis` in `kuma-demo-migration` namespace:
+You can go to {{site.mesh_product_name}} GUI (port 5681) Data Plane Proxies `Stats` section on `redis` in `kuma-demo-migration` namespace, and you should see this metric increment :
```
Automaat, line 98: can we have a screenshot from gui to show exactly where to look for this stats?

app/_src/guides/progressively-rolling-in-strict-mtls.md
```diff
@@ -5,7 +5,42 @@ title: Progressively rolling in strict mTLS
 The [MeshTLS](/docs/{{ page.version }}/policies/meshtls/) policy allows you to gradually migrate services to mutual TLS without dropping a packet.
 
 ## Prerequisites
-- Completed [quickstart](/docs/{{ page.version }}/quickstart/kubernetes-demo/) to set up a zone control plane with demo application
+- Completed [quickstart](/docs/{{ page.version }}/quickstart/kubernetes-demo/) to set up a zone control plane with demo application.
+
+## Basic setup
+
+In order to be able to fully utilize MeshTLS policy you need to enable [Mutual TLS](/docs/{{ page.version }}/policies/mutual-tls/) (mTLS), and we can do it with `builtin` CA backend by executing:
```
slonka, line 12: I thought that we're not supposed to use `we` to indicate `us` and the `user` rather you should use `you`. Do I remember it correctly?

```suggestion
In order to be able to fully utilize MeshTLS policy you need to enable [Mutual TLS](/docs/{{ page.version }}/policies/mutual-tls/) (mTLS), and you can do it with `builtin` CA backend by executing:
```

app/_src/guides/progressively-rolling-in-strict-mtls.md
```diff
@@ -175,13 +219,15 @@ flowchart LR
...
 Finally, to set strict mode you can either edit the policy or remove it (the default is taken from `Mesh` object which is `STRICT`).
 
-```bash
-kubectl delete meshtlses.kuma.io -n kuma-system redis
-```
+{% tip %}
+**Things to remember when migrating to strict TLS**
 
-## Things to remember when migrating to strict TLS
+The difference between `cluster.localhost_6379.upstream_cx_total` and `inbound_10_42_0_13_6379.rbac.allowed` before changing a workload to `Strict` mode equals that after changing it.
```
slonka, line 225: This seems more correct to me, but would love input from a native speaker - @michaelbeaumont - can you take a look?

```suggestion
If there is only encrypted traffic sent to the destination the difference between `cluster.localhost_6379.upstream_cx_total` and `inbound_10_42_0_13_6379.rbac.allowed` before changing a workload to `Strict` mode will equal that after changing it.
```

app/_src/guides/progressively-rolling-in-strict-mtls.md
Icarus9913: In this documentation, I found that `we` and `you` for several times. Are there any norms for this?

app/_src/guides/progressively-rolling-in-strict-mtls.md
michaelbeaumont: My suggestion:

```suggestion
If only encrypted traffic is sent to the destination, the difference between `cluster.localhost_6379.upstream_cx_total` and `inbound_10_42_0_13_6379.rbac.allowed`  will not change after setting the workload to `Strict` mode.
```

"that after changing it" isn't right and "the difference doesn't change" sounds better than "the difference equals .."

app/_src/guides/progressively-rolling-in-strict-mtls.md
michaelbeaumont: Yeah. I don't know. There's a Vale rule that warns you about it. Both sound weird sometimes but probably "you" is better. Kubernetes uses "you".

app/_src/guides/progressively-rolling-in-strict-mtls.md
```diff
@@ -5,7 +5,42 @@ title: Progressively rolling in strict mTLS
 The [MeshTLS](/docs/{{ page.version }}/policies/meshtls/) policy allows you to gradually migrate services to mutual TLS without dropping a packet.
 
 ## Prerequisites
-- Completed [quickstart](/docs/{{ page.version }}/quickstart/kubernetes-demo/) to set up a zone control plane with demo application
+- Completed [quickstart](/docs/{{ page.version }}/quickstart/kubernetes-demo/) to set up a zone control plane with demo application.
+
+## Basic setup
+
+In order to be able to fully utilize MeshTLS policy you need to enable [Mutual TLS](/docs/{{ page.version }}/policies/mutual-tls/) (mTLS), and you can do it with `builtin` CA backend by executing:
```
michaelbeaumont, line 12: Add a line

```
mTLS
```

alphabetically to `.github/styles/config/vocabularies/Base/accept.txt`

app/_src/guides/progressively-rolling-in-strict-mtls.md
```diff
@@ -94,7 +130,11 @@ kubectl patch deployment redis -n kuma-demo-migration \
 ```
 
 After this redis will be receiving plaintext traffic from non-meshed client.
-You can go to {{site.mesh_product_name}} GUI (port 5681) and you should see this metric increment on `redis` in `kuma-demo-migration` namespace:
+You can go to {{site.mesh_product_name}} GUI (port 5681) Data Plane Proxies `Stats` section on `redis` in `kuma-demo-migration` namespace, and you should see this metric increment :
```
michaelbeaumont, line 133:
```suggestion
You can go to the {{site.mesh_product_name}} GUI, check the `Stats` tab for the `redis` Dataplane in the `kuma-demo-migration` namespace, and you should see this metric increment:
```

PR #1971, opened 2024-10-15
docs(install): replace Kind with Minikube for simple test installations

### LINKED ISSUE #1649: Suggest using minikube for guides
### Description

For guides minkube is a little nicer because it has tunnels etc.

Let's update the guides to suggest this instead of kind etc.

dascole:
- Updated installation instructions to use Minikube instead of Kind.
- Modified references throughout relevant pages to reflect Minikube.
- Ensured steps are clear for users to set up Minikube.

Closes #1649

Signed-off-by: dascole joe.dascole@konghq.com

### DIFF
--- a/.github/styles/config/vocabularies/Base/accept.txt
+++ b/.github/styles/config/vocabularies/Base/accept.txt
@@ -95,6 +95,7 @@ Mockbin
 Moesif
 MongoDB
 Multus
+mTLS
 NAD
 namespace[sd]?
 navtabs
@@ -138,7 +139,6 @@ target[Rr]ef
 tbl
 tcpdump
 TLS
-tls
 transcoder
 ttl
 txt
--- a/app/_src/guides/federate.md
+++ b/app/_src/guides/federate.md
@@ -28,10 +28,10 @@ minikube start -p mesh-global
 ```
 
 ### Setup the minikube tunnel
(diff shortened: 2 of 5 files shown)


dascole:
Setting this PR to draft because the IP addresses haven't been updated yet to reflect Minikube's tunnel connection. Will mark it ready for review once the IPs are updated.

.github/styles/config/vocabularies/Base/accept.txt
```diff
@@ -95,6 +95,7 @@ Mockbin
 Moesif
 MongoDB
 Multus
+mTLS
```
michaelbeaumont, line 98: Argh this is in multiple PRs sorry. We need one entry I think

```
m?(TLS|tls)
```

Though I guess remove lowercase if you want

michaelbeaumont (approved):

.github/styles/config/vocabularies/Base/accept.txt
lahabana: I'd remove lowercase because we only allow TLS. Coherency is just easier :)

michaelbeaumont: yeah i meant more important is the `m?TLS` part otherwise I think it will mismatch mTLS as TLS depending on which one comes first. At least it seemed like that in https://github.com/kumahq/kuma-website/pull/1967 but we can also add it in a separate PR

PR #1977, opened 2024-10-17
ci(vale): reject controlplane/control-plane and dataplane/data-plane

michaelbeaumont:
We want to write control plane and data plane, Dataplane is still allowed.


lahabana (approved):
Thx for doing this I never know how to write this

PR #1984, opened 2024-10-17
ci(vale): fix reporter, only annotate errors

michaelbeaumont:
This will only show errors as annotations on the diff.

We are currently not showing anything from Vale. This will show errors. For example, rejected word errors


Automaat:
Why are we changing this? Can we add more context to pr description?

michaelbeaumont:
> Why are we changing this? Can we add more context to pr description?

What's missing from the description? We weren't showing annotations. This adds errors as annotations.

lukidzi (approved):

PR #1988, opened 2024-10-18
ci(vale): ignore mermaid blocks

michaelbeaumont:
Ideally we'd check _some_ of the contents but there's too much noise since Vale doesn't natively understand it.


Automaat (approved):

Automaat:
I think we should just start with allowing certain words from mermaid syntax. Because of this we can benefit from using vale to check if names on diagrams are consistent with our naming policy

michaelbeaumont:
Closing for https://github.com/kumahq/kuma-website/pull/1989

PR #1996, opened 2024-10-21
Add meshservice checkbox to tabs component example

slonka:
just to show @johnharris85 how it looks like


jekyll-kuma-plugins/lib/jekyll/kuma-plugins/liquid/tags/policyyaml.rb
```diff
@@ -2,26 +2,55 @@
...
+            if hash.dig("spec", "to").is_a?(Array)
+              hash["spec"]["to"].each do |to_item|
+                if to_item.dig("targetRef", "kind") == "MeshService"
+                  target_ref = to_item["targetRef"]
+                  if target_ref["name_kube"]
+                    transformed_name = target_ref["name_kube"].split('_')
+                    to_item["targetRef"] = {
+                      "kind" => "MeshService",
+                      "name" => transformed_name[0],
+                      "namespace" => transformed_name[1],
```
Automaat, line 36: we shouldn't have namespace for universal policies, and this code will add it. We should check if this is for universal or k8s

slonka: we talked offline and it will be addressed in #2003

Issue #2020, opened 2024-10-24
Rewrite places not using policy_yaml to use MeshService style to targetRef


slonka:
### Description

The https://github.com/kumahq/kuma-website/issues/1993 covers places that use `policy_yaml` but we have other places like https://github.com/kumahq/kuma-website/blame/0471523aa41cb7856e71114504ed31d90d389509/app/_src/policies/meshtcproute.md#L71 that need changing. I'll mark these with TODO and link to this issue.

slonka:
Marcin kindly volunteered to take this one.

PR #2052, opened 2024-10-30
feat(policies): move policies examples to policy_yaml

Automaat:
Fixes: #2020


app/_src/policies/meshcircuitbreaker.md
```diff
@@ -474,7 +474,35 @@ spec:
 
 #### Basic circuit breaker for outbound traffic from web, to backend service
 
-{% policy_yaml usage use_meshservice=true %}
+{% if_version lte:2.8.x %}
+{% policy_yaml usage-28x use_meshservice=true %}
```
lahabana, line 478: If this is lte 2.8 shoudn't it not be using meshservice?

lahabana (approved):
Thank you so much!

app/_src/policies/meshcircuitbreaker.md
Automaat: removed

app/_src/policies/meshhttproute.md
```diff
@@ -462,15 +478,17 @@ spec:
                 requestMirror:
                   percentage: 30
                   backendRef:
-                    kind: MeshServiceSubset
-                    name: backend_kuma-demo_svc_3001
-                    tags:
-                      version: v1_experimental
+                    kind: MeshService
```
Automaat, line 481: I am not sure what to do with this, the example generated from it when not using MeshService in incorrect. There is the same example in MeshTCPRoute. For old scenario with `kuma.io/service` we should use MeshServiceSubset and with tags

slonka: We should fix the generator code right? Can we if_version it for now, create an issue to fix it and merge this?

slonka (approved):

app/_src/policies/meshhttproute.md
slonka: oh, it's in `backendRef` - maybe @lobkovilya has an idea how to fix it

slonka: fixed this on https://github.com/kumahq/kuma-website/pull/2056

Automaat: it was fixed with https://github.com/kumahq/kuma-website/pull/2056 I will update examples

PR #2056, opened 2024-10-31
fix(policyyaml) fix policy yaml backendref tags

slonka:
Fix backendref with tags and mirror

### DIFF
--- a/spec/fixtures/mhr-port.yaml
+++ b/spec/fixtures/mhr-port.yaml
@@ -26,8 +26,8 @@ spec:
                   percentage: 30
                   backendRef:
                     kind: MeshService
-                    name: backend_experimental
-                    _version: v1
+                    name: backend
+                    _version: experimental
                     namespace: kuma-demo
                     port: 3001
             backendRefs:
--- a/spec/fixtures/mhr-port.yaml
+++ b/spec/fixtures/mhr-port.yaml
@@ -1,0 +1,38 @@
+type: MeshHTTPRoute
+name: http-route-1
+mesh: default
+spec:
+  targetRef:
+    kind: MeshSubset
+    tags:
+      app: frontend
+  to:
+    - targetRef:
+        kind: MeshService
+        name: backend
+        namespace: kuma-demo
+        sectionName: http
+        _port: 3001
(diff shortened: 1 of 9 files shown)


spec/fixtures/mtr.yaml
```diff
@@ -0,0 +1,31 @@
...
+        _port: 3001
+      rules:
+        - default:
+            backendRefs:
+              - kind: MeshService
+                name: backend
+                _version: v1
+                namespace: kuma-demo
+                _port: 3001
+                port: 3001
```
Automaat, line 23: do we need `port` in this example?

spec/fixtures/mhr-port.yaml
```diff
@@ -0,0 +1,38 @@
...
+                  name: mirror-this-request
+                  value: "true"
+          default:
+            filters:
+              - type: RequestMirror
+                requestMirror:
+                  percentage: 30
+                  backendRef:
+                    kind: MeshService
+                    name: backend_v1_experimental
```
Automaat, line 29: i think this should use `_version` instead of full name

slonka:
[localhost:8000.zip](https://github.com/user-attachments/files/17588078/localhost.8000.zip)

```
python -m http.server 8000
```

open http://localhost:8000/localhost%3A8000/spec/fixtures/mhr-and-mtr.golden.html

slonka:
new rendered:

[localhost:8000.zip](https://github.com/user-attachments/files/17588208/localhost.8000.zip)

spec/fixtures/mtr.yaml
slonka: we only need `port`, not `_port`

Automaat: so we can remove it from this example

Automaat: so we now don't use _port in our examples? won't it break all of the rest already migrated policies?

slonka: `_port` in `targetRef` is different than `port` in `backendRef`

Automaat (approved):

PR #2060, opened 2024-11-04
fix(guide): fix and move guide for Kubernetes Gateway API

Automaat:
Old guide is not working. I have fixed it and I would like to move it to guides section

### DIFF
--- a/app/_src/guides/gateway-api.md
+++ b/app/_src/guides/gateway-api.md
@@ -1,0 +1,336 @@
+---
+title: Kubernetes Gateway API
+---
+
+To get traffic from outside your mesh inside it (North/South) with {{site.mesh_product_name}} you can use
+a builtin gateway.
+
+In the [quickstart](/docs/{{ page.version }}/quickstart/kubernetes-demo/), traffic was only able to get in the mesh by port-forwarding to an instance of an app
+inside the mesh.
+In production, you typically set up a gateway to receive traffic external to the mesh.
+In this guide you will add [a built-in gateway](/docs/{{ page.version }}/using-mesh/managing-ingress-traffic/builtin/) in front of the demo-app service and expose it publicly.
+We will deploy and configure Gateway using [Kubernetes Gateway API](https://gateway-api.sigs.k8s.io/).
+
+{% mermaid %}
+---
+title: service graph of the demo app with a builtin gateway on front
+---
+flowchart LR
+  subgraph edge-gateway
+    gw0(/ :8080)
+  end
+  demo-app(demo-app :5000)
+  redis(redis :6379)
+  gw0 --> demo-app
+  demo-app --> redis
+{% endmermaid %}
+
+## Prerequisites
+
+- Completed [quickstart](/docs/{{ page.version }}/quickstart/kubernetes-demo/) to set up a zone control plane with demo application
+
+## Install Gateway API CRDs
+
+To install Gateway API please refer to [official installation instruction](https://gateway-api.sigs.k8s.io/guides/#install-standard-channel).
+
+You also need to manually install {{site.mesh_product_name}} [GatewayClass](https://gateway-api.sigs.k8s.io/api-types/gatewayclass/):
+
+```shell
+echo "apiVersion: gateway.networking.k8s.io/v1
+kind: GatewayClass
+metadata:
+  name: kuma
+spec:
+  controllerName: gateways.kuma.io/controller" | kubectl install -f -
+```
+
+At this moment, when you install Gateway API CRDs after installing {{site.mesh_product_name}} control plane you need to restart
+it to start Gateway API controller. To do this run: 
+
+```shell
+kubectl rollout restart deployment {{site.mesh_product_name_path}}-control-plane -n {{ site.mesh_namespace }}
+```
+
+## Start a gateway
+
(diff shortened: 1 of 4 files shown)


Automaat:
https://deploy-preview-2060--kuma.netlify.app/docs/2.9.x/guides/gateway-api/

app/_src/guides/gateway-api.md
```diff
@@ -0,0 +1,336 @@
...
+
+## Prerequisites
+
+- Completed [quickstart](/docs/{{ page.version }}/quickstart/kubernetes-demo/) to set up a zone control plane with demo application
+
+## Install Gateway API CRDs
+
+To install Gateway API please refer to [official installation instruction](https://gateway-api.sigs.k8s.io/guides/#install-standard-channel).
+
+You also need to manually install {{site.mesh_product_name}} [GatewayClass](https://gateway-api.sigs.k8s.io/api-types/gatewayclass/):
```
michaelbeaumont, line 34: this isn't true if you install the CRDs first

app/_src/guides/gateway-api.md
```diff
@@ -0,0 +1,336 @@
...
+
+```shell
+echo "apiVersion: gateway.networking.k8s.io/v1
+kind: GatewayClass
+metadata:
+  name: kuma
+spec:
+  controllerName: gateways.kuma.io/controller" | kubectl install -f -
```
michaelbeaumont, line 42:
```suggestion
  controllerName: gateways.kuma.io/controller" | kubectl apply -f -
```

app/_src/guides/gateway-api.md
```diff
@@ -0,0 +1,336 @@
...
+spec:
+  gatewayClassName: kuma
+  listeners:
+   - name: proxy
+     port: 8080
+     protocol: HTTP" | kubectl apply -f -
+```
+
+{% warning %}
+The Kubernetes cluster needs to support `LoadBalancer` for this to work.
```
michaelbeaumont, line 72: alternatively we can could create a GatewayClass + MeshGatewayConfig that uses ClusterIP

app/_src/guides/gateway-api.md
```diff
@@ -0,0 +1,336 @@
...
+
+If you are running `minikube` you will want to open a [tunnel](https://minikube.sigs.k8s.io/docs/handbook/accessing/#loadbalancer-access) with `minikube tunnel -p mesh-zone`.
+
+You may not have support for `LoadBalancer` if you are running locally with `kind` or `k3d`.
+One option for `kind` is [kubernetes-sigs/cloud-provider-kind](https://github.com/kubernetes-sigs/cloud-provider-kind) may be helpful.
```
michaelbeaumont, line 77:
```suggestion
One option for `kind` is [kubernetes-sigs/cloud-provider-kind](https://github.com/kubernetes-sigs/cloud-provider-kind).
```

app/_src/guides/gateway-api.md
```diff
@@ -0,0 +1,336 @@
...
+{% endwarning %}
+
+You can now check if gateway is running in demo app `kuma-demo` namespace:
```
michaelbeaumont, line 80:
```suggestion
You can now check if the gateway is running in the demo app `kuma-demo` namespace:
```

app/_src/guides/gateway-api.md
```diff
@@ -0,0 +1,336 @@
...
+< date: Fri, 09 Feb 2024 10:10:16 GMT
+< server: Kuma Gateway
+< x-envoy-upstream-service-time: 24
+<
+* Connection #0 to host 127.0.0.1 left intact
+RBAC: access denied%
+```
+
+Notice the forbidden error.
+This is because the quickstart has very restrictive permissions as defaults.
```
michaelbeaumont, line 179: Does it? Does that mean `kumactl install demo` doesn't work?

app/_src/using-mesh/managing-ingress-traffic/gateway-api.md
```diff
@@ -9,6 +9,11 @@ for configuring {% if_version gte:2.6.x inline:true %}[built-in gateway](/docs/{
 
 ## Installation
 
+{% if_version gte:2.9.x %}
+Checkout dedicated installation [guide](/docs/{{ page.version }}/guides/gateway-api).
```
michaelbeaumont, line 13:
```suggestion
Checkout our dedicated installation [guide](/docs/{{ page.version }}/guides/gateway-api).
```

app/_src/guides/gateway-api.md
Automaat: yes, but this guide assumes that user followed quickstart and already have control plane and demo installed

app/_src/guides/gateway-api.md
Automaat: `kumactl install demo` does work, but in quickstart we configure mTLS and create MTP that only allows traffic from demo-app to redis that is why it will be forbidden here, this is the same as in other gateway guides

app/_src/guides/gateway-api.md
Automaat: do we need an issue for this? If yes, can you create one with better explanation? I don't want to miss any context

michaelbeaumont: Probably this is good as is

michaelbeaumont (approved):

PR #2061, opened 2024-11-04
fix(guides): update gateways guides and stop using topLevel MeshService targetRef in our examples

Automaat:
We should not use MeshService in topLevel targetRef since it became deprecated in 2.9.0


lahabana (approved):
Needs some rebase but LGTM

PR #2062, opened 2024-11-04
remove mesh metric otel from experimental

johnharris85:
Remove the `experimental` designation for OTEL in MeshMetric.


Automaat (approved):

Automaat (changes requested):
please add `{% if_version gte:2.9.x %}` for this change so that we remove experimental only form 2.9.x version

johnharris85:
Is 2.9 correct? Further down there is:
```
{% if_version gte:2.7.x %}
#### OpenTelemetry
...
```

Removing the 'experimental' designation from 2.7? Assuming that's correct?

Automaat:
@johnharris85 do we have `experimental` somewhere in nav bar? I cannot find it. I see that you put `gte 2.7` I think we should remove experimental only from the newest version. It was experimental in 2.7, and it had some bugs. Also, it will motivate users to use the newest version

Automaat (approved):

PR #2069, opened 2024-11-06
feat(guides): move guides to use new MeshService by default

Automaat:
We've discussed that for version newer than 2.9.x We should push new users to use new MeshService. This pr updated and fixes guides to work with MeshService mode `Exclusive`

### DIFF
--- a/app/_src/guides/consumer-producer-policies.md
+++ b/app/_src/guides/consumer-producer-policies.md
@@ -15,30 +15,13 @@ This guide will help you get comfortable with producer consumer model.
 ## Basic setup
 
 In order to be able to fully utilize namespace scoped policies you need to use [MeshService](/docs/{{ page.version }}/networking/meshservice). 
-To enable MeshService resource generation on Mesh level we need to run:
-
-```shell
-echo "apiVersion: kuma.io/v1alpha1
-kind: Mesh
-metadata:
-  name: default
-spec:
-  meshServices:
-    mode: Exclusive
-  mtls:
-    enabledBackend: ca-1
-    backends:
-    - name: ca-1
-      type: builtin" | kubectl apply -f -
-```
-
 To make sure that traffic works in our examples let's configure MeshTrafficPermission to allow all traffic:
 
 ```shell
 echo "apiVersion: kuma.io/v1alpha1
 kind: MeshTrafficPermission
 metadata:
-  namespace: kuma-system
+  namespace: kuma-demo
   name: mtp
 spec:
   targetRef:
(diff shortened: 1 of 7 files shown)


app/_src/quickstart/universal-demo.md
```diff
@@ -198,6 +199,20 @@ mtls:
     - name: ca-1
       type: builtin' | kumactl apply -f -
 ```
+{% endif_version %}
+{% if_version gte:2.9.x %}
+```sh
+echo 'type: Mesh
+name: default
+meshServices:
+  mode: Exclusive
```
lahabana, line 208: Have you checked this is all that's needed? I think there's no auto meshservice generator by default in universal

app/_src/guides/gateway-builtin.md
```diff
@@ -291,17 +334,16 @@ spec:
...
         tags:
-          port: http-8080
-" | kubectl apply -f -
+          port: http-8080" | kubectl apply -f -
 ```
 
 Check the call to the gateway: 
-```shell
+```sh
 curl -X POST -v --insecure https://${PROXY_IP}:8080/increment
```
lahabana, line 342:
```suggestion
curl -X POST -v --insecure "https://${PROXY_IP}:8080/increment"
```

For some reason for this one the copy paste adds backslash. I think the quotes should avoid this

lahabana (approved):
Tested the getting started and the builtin gateway guides and it worked well!
Thanks for improving this!

app/_src/guides/consumer-producer-policies.md
```diff
@@ -15,30 +15,13 @@ This guide will help you get comfortable with producer consumer model.
 ## Basic setup
 
 In order to be able to fully utilize namespace scoped policies you need to use [MeshService](/docs/{{ page.version }}/networking/meshservice).
```
lahabana, line 17: Should we just remove this line and just consider the user as using `MeshService` ?

app/_src/quickstart/universal-demo.md
Automaat: It does not look like we need anything according to docs: https://kuma.io/docs/2.9.x/networking/meshservice/#universal Also I check and I see MeshService generated after starting dp

app/_src/guides/consumer-producer-policies.md
Automaat: I was thinking about this, but I don't think we can assume users are using MeshService, this guide is linked in blog post for release, and there can be users coming to check out new feature with already existing environments. Without starting with new quickstart

PR #2070, opened 2024-11-07
feat(guides): update otel collector guide to work with newest otel collector

Automaat:
Our guide was outdated, and caused issues with otel

### DIFF
--- a/app/_src/guides/otel-metrics.md
+++ b/app/_src/guides/otel-metrics.md
@@ -34,10 +34,8 @@ config:
   extensions:
     health_check:
       endpoint: \${env:MY_POD_IP}:13133
-    memory_ballast: {}
   processors:
     batch: {}
-    memory_limiter: null
   receivers:
     otlp:
       protocols:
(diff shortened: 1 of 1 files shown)


app/_src/guides/otel-metrics.md
```diff
@@ -144,9 +145,26 @@ spec:
     backends:
       - type: OpenTelemetry
         openTelemetry:
-          endpoint: opentelemetry-collector.mesh-observability.svc:4317
-' | kubectl apply -f -
+          endpoint: opentelemetry-collector.mesh-observability.svc:4317' | kubectl apply -f -
 ```
+{% endif_version %}
+{% if_version gte:2.9.x %}
```
michaelbeaumont, line 151: I think we don't need the two if_versions, we can fix it for all versions right?

michaelbeaumont (approved):

app/_src/guides/otel-metrics.md
Automaat: The policy is slightly different in 2.9.x, we are using syntactic sugar to omit topLevel targetRef

michaelbeaumont: Ah sorry ok I thought it was the newline at the beginning and end

PR #2082, opened 2024-11-12
feat(prod/cp-deployment): add hint to restart pods after installation

schogges:
After running into this, I think it might be worth to document this.


app/_src/production/cp-deployment/single-zone.md
```diff
@@ -7,7 +7,7 @@ In order to deploy {{site.mesh_product_name}} in a single-zone deployment, the `
 
 {% tabs usage useUrlFragment=false %}
 {% tab usage Kubernetes %}
-This is the standard installation method. 
+This is the standard installation method. Make sure to restart service pods that are already running after installation such that the data plane proxies can be connected.
```
jijiechen, line 10:
```suggestion
This is the standard installation method. After the installation, make sure to restart application pods that are  running such that the data plane proxies can be generated and connected.
```

jijiechen: Here we are assuming the application could have started before the mesh installation. 

I personally feel this is not a practical scenario. Applications need to know whether they need mesh and whether mesh should be enabled on them before they go live. 

That been said, apps that require mesh should not be running before the CP. If they were running for some reason, people should have known they need to restart them.

jijiechen: Well, feel free to resolve this conversation. From a test user's perspective, apps and mesh CP can be started in any order. So it's useful for beginers.

app/_src/production/cp-deployment/multi-zone.md
```diff
@@ -216,6 +216,8 @@ Set `--set {{site.set_flag_values_prefix}}controlPlane.tls.kdsZoneClient.skipVer
 For production use a certificate signed by a trusted CA. See [Secure access across services](/docs/{{ page.release }}/production/secure-deployment/certificates/) page for more information.
 {% endif_version %}
 
+After installing a zone control plane, make sure to restart the service pods that are already running such that the data plane proxies can be connected.
```
jijiechen, line 219:
```suggestion
After installing a zone control plane, make sure to restart the application pods that are already running such that the data plane proxies can be connected.
```

jijiechen:
Looks good to me basically. Please have a look at the suggestions.

app/_src/production/cp-deployment/single-zone.md
schogges: Yes, I agree that most users know about this 🙂 but as me as a beginner ran into this, I thought it could help.

michaelbeaumont (approved):

jijiechen (approved):

PR #2116, opened 2024-12-10
ci(checklinks): exclude external website from checklinks

Icarus9913:
I got some checklinks failure but it did works from my PC:

```text
https://deploy-preview-2114--kuma.netlify.app/docs/2.9.x/production/dp-config/cni/
	403	https://docs.cilium.io/en/v1.14/operations/upgrade/#id2
https://deploy-preview-2114--kuma.netlify.app/docs/2.9.x/production/cp-deployment/kubernetes/
	403	https://argo-cd.readthedocs.io/en/stable/
https://deploy-preview-2114--kuma.netlify.app/docs/2.7.x/production/dp-config/cni/
	403	https://docs.cilium.io/en/v1.14/operations/upgrade/#id2
https://deploy-preview-2114--kuma.netlify.app/docs/2.7.x/production/cp-deployment/kubernetes/
	403	https://argo-cd.readthedocs.io/en/stable/
https://deploy-preview-2114--kuma.netlify.app/docs/2.8.x/production/dp-config/cni/
	403	https://docs.cilium.io/en/v1.14/operations/upgrade/#id2
https://deploy-preview-2114--kuma.netlify.app/docs/2.8.x/production/cp-deployment/kubernetes/
	403	https://argo-cd.readthedocs.io/en/stable/
Error: Process completed with exit code 1.
```

Did you sign your commit? [Instructions](https://github.com/kumahq/.github/blob/main/CONTRIBUTING.md#sign-your-commits)

Have you read [Contributing guidelines](https://github.com/kumahq/.github/blob/main/CONTRIBUTING.md)?

PR #2118, opened 2024-12-10
Revert "ci(checklinks): exclude external website from checklinks"

slonka:
Reverts kumahq/kuma-website#2116

As explained in the review, this is not the way. If we go down this road we might as well disable all external links checking.


Automaat:
closing as we no exclude all external links from checking

PR #2126, opened 2024-12-19
feat(mes): add example of disabling traffic to MeshExternalService on Mesh

Automaat:
I've simplified some parts of MES docs, and added an example on how to disable traffic to MES

### DIFF
--- a/.github/styles/config/vocabularies/Base/accept.txt
+++ b/.github/styles/config/vocabularies/Base/accept.txt
@@ -180,3 +180,4 @@ yml
 [xX]DS|xds
 Zipkin
 zsh
+MeshExternalService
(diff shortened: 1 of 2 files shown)


lukidzi (approved):

.github/styles/config/vocabularies/Base/accept.txt
```diff
@@ -180,3 +180,4 @@ yml
 [xX]DS|xds
 Zipkin
 zsh
+MeshExternalService
```
slonka, line 183: this list is alphabetical, please put it in the right place

slonka (approved):

.github/styles/config/vocabularies/Base/accept.txt
Automaat: fixed

PR #2149, opened 2025-01-10
docs(MeshTrafficPermission): Add a note to enable mTLS to make mtp works

Icarus9913:
When I went through the MeshTrafficPermission source codes and found we need to enable the mTLS then the MTP policy works.

https://github.com/kumahq/kuma/blob/2.9.2/pkg/plugins/policies/meshtrafficpermission/plugin/v1alpha1/plugin.go#L51-L56

### DIFF
--- a/app/_src/policies/meshtrafficpermission.md
+++ b/app/_src/policies/meshtrafficpermission.md
@@ -8,7 +8,7 @@ Do **not** combine with [TrafficPermission](/docs/{{ page.release }}/policies/tr
 {% endwarning %}
 
 {% tip %}
-Remember [Mutual TLS](/docs/{{ page.release }}/policies/mutual-tls) has to be enabled to make MeshTrafficPermission works.
+[Mutual TLS](/docs/{{ page.release }}/policies/mutual-tls) has to be enabled to make MeshTrafficPermission work.
 {% endtip %}
 
 The `MeshTrafficPermission` policy provides access control within the [Mesh](/docs/{{ page.release }}/production/mesh/).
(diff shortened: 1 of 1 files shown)


jakubdyszkiewicz:
Make sense to add it, but why in the warning? I think MeshTrafficPermission is just missing description of what it is, which would include requirement on mTLS. See

image
image

jakubdyszkiewicz (commented):
see comment

Icarus9913:
Yeah, I gonna supplement the description and move that note into it

Icarus9913:
@jakubdyszkiewicz I added the description and moved the mTLS requirement into a **Tip**, I guess the Tip might be more appropriate.

app/_src/policies/meshtrafficpermission.md
```diff
@@ -7,6 +7,13 @@ This policy uses new policy matching algorithm.
 Do **not** combine with [TrafficPermission](/docs/{{ page.release }}/policies/traffic-permissions).
 {% endwarning %}
 
+{% tip %}
+Remember [Mutual TLS](/docs/{{ page.release }}/policies/mutual-tls) has to be enabled to make MeshTrafficPermission works.
```
jakubdyszkiewicz, line 11: "Remember Mutual TLS has to be enabled to make MeshTrafficPermission works" sounds kind of weird. Sounds like it's missing comma or "remember" is redundant. How about 
```
mTLS has to be enabled to make MeshTrafficPermission work.
```
or
```
MeshTrafficPermission requires enabled mTLS in the Mesh.
```

PR #2201, opened 2025-02-27
feat(dataplane): use dataplane kind in policy examples

Automaat:
We've added new kind Dataplane we should use it in policies examples in favour of deprecated MeshSubset

### DIFF
--- a/app/_src/policies/meshcircuitbreaker.md
+++ b/app/_src/policies/meshcircuitbreaker.md
@@ -512,7 +512,8 @@ spec:
 ```
 {% endpolicy_yaml %}
 {% endif_version %}
-{% if_version gte:2.9.x %}
+
+{% if_version eq:2.9.x %}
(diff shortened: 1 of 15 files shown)


app/_src/policies/meshcircuitbreaker.md
```diff
@@ -591,6 +621,46 @@ spec:
...
+{% policy_yaml protocol-210x namespace=kuma-demo %}
+```yaml
+type: MeshCircuitBreaker
+name: backend-inbound-outlier-detection
+mesh: default
+spec:
+  targetRef:
+    kind: Dataplane
+    labels:
+      app: web
```
lukidzi, line 663: should we use rules here?

Automaat: I will, but I started with just changing docs for dataplane kind. We have separate issues for `rules` api

Automaat:
@lobkovilya I've just merged it: https://github.com/kumahq/kuma-website/pull/2203

Issue #2137, opened 2024-12-23
Update Policies/Introduction with the new `spec.rules` field


lobkovilya:
### Description

https://kuma.io/docs/2.9.x/policies/introduction/ should reflect the [new design](https://docs.google.com/document/d/1tdIOVVYObHbGKX1AFhbPH3ZQKYvpQG-2s6ShHSECaXM/edit?tab=t.0#heading=h.92b1f3be2g7w)

PR #2215, opened 2025-03-06
docs(policy): update introduction.md with `rules` and new kind `Dataplane`

lobkovilya:
Fix https://github.com/kumahq/kuma-website/issues/2137


bartsmykla (changes requested):
tabs needs to be adjusted here after: https://github.com/kumahq/kuma-website/pull/2216

app/_src/policies/introduction.md
```diff
@@ -64,14 +64,63 @@ The `spec` field contains the actual configuration of the policy.
...
+        kind: MeshService # only for requests destined for 'my-service'
+        name: my-service
+      default: # configuration that applies to selected requests on selected proxies
+        key: value
+---
+type: ExampleInboundPolicy
+name: my-example
+mesh: default
+spec:
+  targetRef:
```
Automaat, line 109: I am not sure if we should introduce Dataplane kind here, as it can be also used for outbound policies. I think we should stick with Mesh

lobkovilya: yeah but the example above `ExampleOutboundPolicy` already shows `targetRef.kind: Mesh`, so I think it makes sense to show all possible variations here

PR #2217, opened 2025-03-07
docs: add guide for upgrading transparent proxy

bartsmykla:
Added a step-by-step guide for upgrading the transparent proxy

---

Preview: https://deploy-preview-2217--kuma.netlify.app/docs/2.9.x/guides/upgrading-transparent-proxy/

---


bartsmykla:
Note that https://github.com/kumahq/kuma-website/pull/2216 affects this PR

app/_src/guides/upgrading-transparent-proxy.md
```diff
@@ -0,0 +1,54 @@
...
+
+{% capture Warning %}{% if page.edition and page.edition != "kuma" %}**Warning:** {% endif %}{% endcapture %}
+{% capture Important %}{% if page.edition and page.edition != "kuma" %}**Important:** {% endif %}{% endcapture %}
+{% capture brbr %}{% if page.edition and page.edition != "kuma" %}<br /><br />{% endif %}{% endcapture %}
+
+The core `iptables` rules applied by {{ Kuma }}'s transparent proxy rarely change, but occasionally new features may require updates. To upgrade the transparent proxy on Universal environments, follow these steps:
+
+### Step 1: Cleanup existing iptables rules{% if_version gte:2.9.x inline:true %} (conditional){% endif_version %}
+
+{% if_version gte:2.9.x %}
```
lobkovilya, line 17: Is it possible to invert it? To have a warning only on docs older than 2.9, and on 2.9 and newer have it clean?

app/_src/guides/upgrading-transparent-proxy.md
```diff
@@ -0,0 +1,54 @@
...
+iptables --table mangle --flush      # Flush all rules in the mangle table (IPv4)
+ip6tables --table mangle --flush     # Flush all rules in the mangle table (IPv6)
+```
+
+### Step 2: Install new transparent proxy
+
+After clearing the `iptables` rules (if necessary), reinstall the transparent proxy by running:
+
+```sh
+kumactl install transparent-proxy [...]
```
lobkovilya, line 51: I'm not sure here, but I was thinking if we're putting this into `guides,` then every command has to be executable so that users can follow along. I think so far, we're following this principle with guides.

app/_src/guides/upgrading-transparent-proxy.md
bartsmykla: but this warning has only sense for 2.9+

lobkovilya: ah ok, I misread it, btw should be `{% if_version gte:2.10.x %}` as you don't need to see the warning on 2.9

bartsmykla: I don't think so, because there might be the rare situation where you want to upgrade from 2.9.x to different version of 2.9 for example

lobkovilya (approved):

Issue #2230, opened 2025-03-11
Add guide for rules API


lahabana:
### Description

This is in 2.10.x let's add a doc.

PR #2242, opened 2025-03-12
feat(guide): add guide for rules api

Automaat:
Fixes: https://github.com/kumahq/kuma-website/issues/2230

### DIFF
--- a/app/_src/guides/rules.md
+++ b/app/_src/guides/rules.md
@@ -1,0 +1,150 @@
+---
+title: Configuring inbound traffic with Rules API
+---
+
+Using `Rules API` combined with targetRef `Dataplane` kind in {{site.mesh_product_name}} you can easily apply configuration to workload
+inbounds. This guide will show you how to configure MeshTimeout policy on data plane inbound and explain how to utilize this API.
+
+## Prerequisites
+
+- Completed [quickstart](/docs/{{ page.release }}/quickstart/kubernetes-demo/) to set up a zone control plane with demo application
+
+## Basic setup
+
+To make sure that traffic works in our examples let's configure MeshTrafficPermission to allow all traffic:
+
+```shell
+echo "apiVersion: kuma.io/v1alpha1
+kind: MeshTrafficPermission
+metadata:
+  namespace: kuma-demo
+  name: mtp
+spec:
+  from:
+    - targetRef:
+        kind: Mesh
+      default:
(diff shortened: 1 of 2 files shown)


app/_src/guides/rules.md
```diff
@@ -0,0 +1,150 @@
...
+As we can see in policy that we've just applied, we use `rules` field to specify configuration for all incoming traffic to our data plane.
+
+```yaml
+rules:
+  - default:
+      http:
+        requestTimeout: 1s
+```
+
+In this scenario we have applied **1 s request timeout** for incoming requests. At this point in time rules api is really simple.
```
lahabana, line 148:
```suggestion
In this scenario we have applied a **request timeout of 1 second** for incoming requests. At this point in time rules api is really simple.
```

lahabana (approved):

app/_src/guides/rules.md
```diff
@@ -0,0 +1,150 @@
...
+You can select data plane by labels:
+
+```yaml
+targetRef:
+  kind: Dataplane
+  labels:
+    app: demo-app
+```
+
+When your application exposes multiple named inbounds. You can select single inbound from your data plane by utilizing `sectionName` field.
```
Icarus9913, line 128: How about using `inbounds(ports)` ?  I guess maybe it's easier for users to understand. WDYT?

app/_src/guides/rules.md
```diff
@@ -0,0 +1,150 @@
...
+```yaml
+rules:
+  - default:
+      http:
+        requestTimeout: 1s
+```
+
+In this scenario we have applied **1 s request timeout** for incoming requests. At this point in time rules api is really simple.
+You cannot apply configuration to subset of incoming traffic. Because of this we don't support
+rules api for MeshTrafficPermission and MeshFaultInjection yet.
```
Icarus9913, line 150: Just one idea when I was reading this section, how about put another subtitle for this paragraph like "Rules API VS From target" ? 

IIUC, the new `Rules API` is a substitute for the specific `from[].targetRef.kind=mesh` right ?

app/_src/guides/rules.md
Automaat: Good catch, but I've handled it differently I've linked to concepts page for inbound in intro to guide, so users at this point should understand what inbound it

Icarus9913 (approved):

PR #2302, opened 2025-05-12
fix(ci): add caddy as a proxy for muffet check link action

Automaat:
Add caddy as a proxy for muffet check link action to avoid GitHub ratelimit by sending authorized requests. This should fix issues with link checking action


Automaat:
no longer needed as github increased limits

PR #2332, opened 2025-06-10
fix(kv-demo): fix links

slonka:
We can use `src` to point to the new source with old links


app/_src/guides/otel-metrics.md
```diff
@@ -18,9 +18,6 @@ lots of possibilities of processing and exporting metrics to multiple ecosystems
 {% endif_version %}
 
 {% if_version gte:2.11.x %}
-- Completed [quickstart](/docs/{{ page.release }}/quickstart/kubernetes-demo-kv/) to set up a zone control plane with demo application
```
Automaat, line 21: this should stay, but we should update link

app/_data/docs_nav_kuma_2.12.x.yml
```diff
@@ -507,11 +508,13 @@ items:
       - text: Producer and consumer policies
         url: /guides/consumer-producer-policies-kv/
       - text: Configuring inbound traffic with Rules API
-        url: /guides/rules/
+        url: /guides/rules-kv/
```
lahabana, line 511: This looks like a url breakage right? Shouldn't this use src as well?

app/_src/guides/otel-metrics.md
slonka: we have a link above, this should've been completely removed, right?

slonka:
```
{% if_version lte:2.10.x %}
```
I mean this ☝️

app/_src/guides/consumer-producer-policies-kv.md
```diff
@@ -14,7 +14,7 @@ This guide will help you become familiar with the model and show how it applies
 
 ## Prerequisites
 
-- Completed [quickstart](/docs/{{ page.release }}/quickstart/kubernetes-demo-kv/) to set up a zone control plane with demo application
+- Completed [quickstart](/docs/{{ page.release }}/quickstart/kubernetes-demo/) to set up a zone control plane with demo application
```
jijiechen, line 17: Without `-kv` in the link, content of this page is actually the same to `app/_src/guides/consumer-producer-policies.md` (the file without `-kv`) right? So we'll also need to remove this file and fix any link pointing to this page accordingly?
Same changes apply to other pages involved.

slonka: We're not changing the link to it - https://github.com/kumahq/kuma-website/pull/2332/files#diff-5c35aa96d3ded97a9857ee2511e5908e44f1ab85ab1d8f7f00f274242c32d422R509 - I assumed it was net new

slonka: it is not net new, we need to revert these links as well

app/_src/guides/otel-metrics.md
Automaat: you need to remove this if `{% if_version lte:2.10.x %}`

PR #2340, opened 2025-06-18
Add targetRef as a concept

justindavies:
Adding targetRef as a concept

### DIFF
--- a/app/_src/introduction/concepts.md
+++ b/app/_src/introduction/concepts.md
@@ -71,3 +71,17 @@ They can enable traffic management, security, observability and traffic reliabil
 
 Policies always have a clear specific area of impact and goal.
 To learn more about [policies checkout the in depth introduction](/docs/{{ page.release }}/policies/introduction).
+
+### Targets (targetRef)
+
+A **Target** (referenced by `targetRef`) is a way to specify exactly which destination resource or workload a policy or routing rule should apply to. Rather than matching all services or inbounds by name or tag, `targetRef` allows fine-grained targeting of:
+
+- A specific **workload instance** (e.g., a single Dataplane proxy)  
+- A **subset** of inbounds/endpoints for a service
+
+#### Why use `targetRef`?
+
+- **Precision**: Apply policies only to a single pod, VM, or data-plane proxy.  
+- **Segmentation**: Differentiate settings for different versions or tiers of the same service.  
+- **Safety**: Roll out policies gradually by targeting one instance at a time.
+


app/_src/introduction/concepts.md
```diff
@@ -71,3 +71,17 @@ They can enable traffic management, security, observability and traffic reliabil
 
 Policies always have a clear specific area of impact and goal.
 To learn more about [policies checkout the in depth introduction](/docs/{{ page.release }}/policies/introduction).
+
+### Targets (targetRef)
+
+A **Target** (referenced by `targetRef`) is a way to specify exactly which destination resource or workload a policy or routing rule should apply to. Rather than matching all services or inbounds by name or tag, `targetRef` allows fine-grained targeting of:
+
+- A specific **workload instance** (e.g., a single Dataplane proxy)  
+- A **subset** of inbounds/endpoints for a service
```
Automaat, line 80: If I understand correctly, you are taking here about destination service. Using name inbound in this case is not the best option as we define `Inbound` as a part of a dataplane proxy. In this case, I think it would be best to call it `service port`.

app/_src/introduction/concepts.md
```diff
@@ -71,3 +71,17 @@ They can enable traffic management, security, observability and traffic reliabil
...
+
+#### Why use `targetRef`?
+
+- **Precision**: Apply policies only to a single pod, VM, or data-plane proxy.
```
Automaat, line 84: what is the difference between pod/vm and data plane here? I think that you cannot have data plane without vm or pod so maybe we can remove it?

app/_src/introduction/concepts.md
```diff
@@ -71,3 +71,17 @@ They can enable traffic management, security, observability and traffic reliabil
 
 Policies always have a clear specific area of impact and goal.
 To learn more about [policies checkout the in depth introduction](/docs/{{ page.release }}/policies/introduction).
+
+### Targets (targetRef)
```
Automaat, line 75: Are you planing on using `targets` as a concept in docs? Maybe we can just call this section `targetRef` as it is commonly used in docs and can be easily referenced

app/_src/introduction/concepts.md
```diff
@@ -71,3 +71,17 @@ They can enable traffic management, security, observability and traffic reliabil
...
+
+A **Target** (referenced by `targetRef`) is a way to specify exactly which destination resource or workload a policy or routing rule should apply to. Rather than matching all services or inbounds by name or tag, `targetRef` allows fine-grained targeting of:
```
Automaat, line 77:
```suggestion
A **Target** (referenced by `targetRef`) is a way to specify exactly which resource a policy or routing rule should apply to. Rather than matching all services or inbounds by name or tag, `targetRef` allows fine-grained targeting of:
```

Workloads are also resources in Kuma. We can also link it to previous section with `Resource` concept.

Automaat, line 77: Isn't matching inbounds by name or tag, already fine-grained targeting allowed by targetRef `sectionName` filed?

Issue #2351, opened 2025-06-30
Document MeshLoadBalancingStrategy with MeshHTTPRoutes


lobkovilya:
### Description

Introduced in https://github.com/kumahq/kuma/pull/13794

PR #2361, opened 2025-07-16
docs(MeshLoadBalancingStrategy): update guide on targeting routes

lobkovilya:
Fix https://github.com/kumahq/kuma-website/issues/2351

### DIFF
--- a/app/_src/guides/targeting-meshhttproutes-in-supported-policies.md
+++ b/app/_src/guides/targeting-meshhttproutes-in-supported-policies.md
@@ -9,7 +9,12 @@ content_type: tutorial
 {% assign kuma-system = site.mesh_namespace | default: "kuma-system" %}
 {% assign kuma-control-plane = kuma | append: "-control-plane" %}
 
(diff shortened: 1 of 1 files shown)


app/_src/guides/targeting-meshhttproutes-in-supported-policies.md
```diff
@@ -392,13 +397,152 @@ curl -XPOST localhost:5050/api/counter
...
+Scale `kv` deployment to 3 replicas:
+
+```bash
+kubectl scale deployment/kv -n kuma-demo --replicas=3
+```
+
+### Switch load balancer type to `RingHash`
+
+By default, the load balancer type is `RoundRobin`.
+The only 2 load balancer types that support hash policies are `RingHash` and `Maglev`.
```
lukidzi: could we add links to docs of these policies ?

Automaat: added

Issue #2365, opened 2025-07-22
Document MeshTrafficPermission with SpiffeId support


lobkovilya:
### Description

* Definitely requires update of https://kuma.io/docs/2.11.x/policies/meshtrafficpermission/.
* Sync with @lukidzi and his MeshIdentity work on a guide as it'll like be a single guide that covers MeshIdentity and MeshTrafficPermission with `spiffeId`

Issue #2395, opened 2025-09-01
Remove TargetRef support matrix


lobkovilya:
### Description

Every policy page has a section

<img width="857" height="377" alt="Image" src="https://github.com/user-attachments/assets/d52b4f22-3ae1-4481-b632-f5845f9d1148" />

But since introduction of MeshServices followed by the policy redesign all the support matrixes have to be similar:

* `spec.targetRef` always support only kinds `Mesh/Dataplane`
* `to[].targetRef` supports resources that have `IsDestination: true` – `MeshService/MeshExternalService/MeshMultiZoneService/MeshHTTPRoute`
* `spec.rules[]` don't have targetRef

**What should we have instead of the support matrix?**

We can create 2 guides:
* Apply policies to your own service (consumer and workload-owner policies)
* Apply policies to your clients (producer policies)

In every policy we will have links to either of these guides (or both). These guides will replace [Producer and Consumer policies](https://kuma.io/docs/2.11.x/guides/consumer-producer-policies/) guide.

lobkovilya:
Make sure to cover https://github.com/kumahq/kuma/issues/13769

PR #2398, opened 2025-09-02
docs(MeshTrafficPermission): rules with SPIFFE-based matches 

lobkovilya:
Fix https://github.com/kumahq/kuma-website/issues/2365


app/_src/policies/meshtrafficpermission_experimental.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Overview
+
+`MeshTrafficPermission` policy defines which clients are allowed or denied access to services inside a mesh based on their identities.
+By default, if no policy is present, all requests are denied.
+
+It enables:
+
+* denying all requests by default, or blocking specific clients (by SPIFFE ID)
+  so that no service owner can override them
```
lukidzi, line 17: Do we explain somewhere who is a ServiceOwner and MeshOperator? Would be beneficial to have it clarified

app/_src/policies/meshtrafficpermission_experimental.md
```diff
@@ -0,0 +1,141 @@
...
+* allowing groups of clients, such as everything in the `observability` namespace,
```
lukidzi, line 18: Should we be here more general and give a few examples with policies below?

app/_src/policies/meshtrafficpermission_experimental.md
```diff
@@ -0,0 +1,141 @@
+---
+title: MeshTrafficPermission
+---
+
+{% tip %}
+[Mutual TLS](/docs/{{ page.release }}/policies/mutual-tls) has to be enabled to make MeshTrafficPermission work.
```
Automaat: I think we should link to MeshIdentity here, not to old stuff

app/_src/policies/meshtrafficpermission_experimental.md
```diff
@@ -0,0 +1,141 @@
...
+  targetRef:
+    kind: Dataplane
+    labels:
+      app: my-app
+  rules:
+    - default:
+        deny:
+          - spiffeId:
+              type: Prefix
+              value: "spiffe://other-mesh.us-east-2.mesh.local/ns/legacy-ns"
```
Automaat: is the cross mesh common example? I don't think so, maybe we should stick with single mesh

app/_src/policies/meshtrafficpermission_experimental.md
```diff
@@ -0,0 +1,141 @@
...
+          - spiffeId:
+              type: Exact
+              value: "spiffe://other-mesh.us-east-2.mesh.local/ns/test/sa/client"
```
Automaat:
```suggestion
              value: "spiffe://other-mesh.us-east-2.mesh.local/ns/frontend/sa/v1"
```
Maybe something like this. It is more verbose, and easier to understand

app/_src/policies/meshtrafficpermission_experimental.md
```diff
@@ -0,0 +1,141 @@
...
+        allow:
+          - spiffeId:
+              type: Prefix
+              value: "spiffe://other-mesh.us-east-2.mesh.local"
+```
+{% endpolicy_yaml %}
+
+With this policy in place, workloads labeled `app: my-app` will reject connections from identities under the `legacy-ns` namespace
+as well as the specific `test/client` identity, while continuing to accept connections from all other identities within the `other-mesh.us-east-2.mesh.local` trust domain.
```
Automaat: this is first appearance of trust domain here, I think should link to MeshTrust docs here

app/_src/policies/meshtrafficpermission_experimental.md
```diff
@@ -0,0 +1,141 @@
...
+## Configuration
+
+`MeshTrafficPermission` policy provides a way to specify 3 lists:
+
+* `deny` list -- list of matches for clients that must always be denied.
+* `allow` list -- list of matchers for clients that are explicitly allowed.
+* `allowWithShadowDeny` list -- list of matchers that are allowed, but also logged as if they were denied.
+Useful for testing a policy to ensure no legitimate clients are denied.
+
+Evaluation rules are:
```
Automaat, line 64: I think we need to describe how rules are merged between policies, here or in policies introduction tab. Since this is different from other policies

app/_src/policies/meshtrafficpermission_experimental.md
```diff
@@ -0,0 +1,141 @@
...
+              value: "spiffe://my-mesh.us-east-2.mesh.local/ns/observability"
+```
+{% endpolicy_yaml %}
+
+This policy allows any client in `observability` namespace to consume any service in `my-mesh`.
+Service Owner can opt-out and deny requests from `observability` if they need to:
+
+{% policy_yaml namespace=backend-ns %}
+```yaml
+type: MeshTrafficPermission
```
Automaat, line 119: can we have examples using kuma-demo apps? It is easier later on to just copy it and slightly tweak.

app/_src/policies/meshtrafficpermission_experimental.md
lobkovilya: Good catch, we need to merge MeshIdentity and MeshTrust first to have proper links in this PR.

app/_src/policies/meshtrafficpermission_experimental.md
lobkovilya: Same, let's wait for MeshTrust docs to be merged

app/_src/policies/meshtrafficpermission_experimental.md
lobkovilya: yeah, maybe it's something that can be added in https://kuma.io/docs/dev/introduction/concepts/, but not in this PR, as we're using these terms in other pages as well, so I haven't introduced them here.

I added an issue https://github.com/kumahq/kuma-website/issues/2403

app/_src/policies/meshtrafficpermission_experimental.md
lobkovilya: I don't know, IMO using `observability` ns gives the idea why would anyone want to make group of clients allowed by default mesh-wide

app/_src/policies/meshtrafficpermission_experimental.md
lobkovilya: service account named `v1` seems a bit odd

app/_src/policies/meshtrafficpermission_experimental.md
lobkovilya: Our demo is not well adopted, all pods are in the same namespace and they all use `default` service account. Examples on this page should just give an idea, policies that actually work with kuma-demo could be found in guides.

app/_data/docs_nav_kuma_2.12.x.yml
```diff
@@ -453,6 +453,15 @@ items:
             url: "/policies/meshtrafficpermission/#configuration"
           - text: Examples
             url: "/policies/meshtrafficpermission/#examples"
+      - text: MeshTrafficPermission (experimental)
```
Automaat, line 460: you need to update 2.13.x nav doc

Automaat (changes requested):

app/_src/policies/meshtrafficpermission_experimental.md
lobkovilya: I'm working on policy introduction rewrite as part of the https://github.com/kumahq/kuma-website/issues/2395 effort.

app/_src/policies/meshtrafficpermission_experimental.md
lobkovilya: added link

Issue #2136, opened 2024-12-23
Document new `spec.rules` field for MeshFaultInjection


lobkovilya:
### Description

https://github.com/kumahq/kuma/issues/12372 adds new field to the policy, we need to update docs

PR #2422, opened 2025-10-02
feat(MeshFaultInjection): docs for rules with matches api

Automaat:
Adding docs for new policy api.

Fixes: https://github.com/kumahq/kuma-website/issues/2136

### DIFF
--- a/app/_src/policies/meshfaultinjection_experimental.md
+++ b/app/_src/policies/meshfaultinjection_experimental.md
@@ -8,15 +8,44 @@ title: MeshFaultInjection
 
 ## Overview
 
-With the MeshFaultInjection policy you can easily test your microservices against resiliency. You can inject failures:
-* abort - returns specific http code
-* delay - adds delay to response
-* response bandwidth - limits network connection speed
+With the MeshFaultInjection policy you can easily test your microservices against resiliency. You can:
+* return specific HTTP error codes to verify error-handling logic.
+* introduce artificial response delays to test timeout and retry behavior.
+* restrict response speed to emulate slow or unstable network connections.
 
 With `matches` api you can select subset of clients for which faults should be injected.
 
 
 ## Configuration
+
+MeshFaultInjection supports three types of faults
+
+### Abort
+
+Abort defines a configuration of not delivering requests to destination service and replacing the responses from destination data plane by
+predefined status code.
+
+- `httpStatus` - HTTP status code which will be returned to source side, has to be in [100 - 599] range
+- `percentage` - a percentage of requests on which abort will be injected, has to be in [0.0 - 100.0] range. If the value is a double number, put it in quotes.
+
+### Delay
+
+Delay defines a configuration of delaying a response from a destination.
+
+- `value` - the duration during which the response will be delayed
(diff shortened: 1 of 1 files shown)


app/_src/policies/meshfaultinjection_experimental.md
```diff
@@ -0,0 +1,165 @@
...
+{% tip %}
+[MeshIdentity](/docs/{{ page.release }}/policies/meshidentity) has to be enabled to make `MeshFaultInjection` with `matches` work.
+{% endtip %}
+
+## Overview
+
+With the MeshFaultInjection policy you can easily test your microservices against resiliency. You can inject failures:
+* abort - returns specific http code
+* delay - adds delay to response
+* response bandwidth - limits network connection speed
```
lukidzi: Instead of using these three terms, which are explained later in the configuration, could we introduce common use cases instead? I think in this section we should explain uses cases generally. Also, we could add some full example

app/_src/policies/meshfaultinjection_experimental.md
```diff
@@ -0,0 +1,165 @@
...
+```yaml
+http:
+  - abort:
+      httpStatus: 500
+      percentage: 70
+  - abort:
+      httpStatus: 503
+      percentage: 50
+```
+That means that for 70% of requests, it returns 500 and for 50% of the 30% that passed it returns 503.
```
lukidzi, line 83: Should we change this to something like: “MeshFaultInjection supports three types of faults" We can list them first, and then add examples in each section where we explain them?

app/_src/policies/meshfaultinjection_experimental.md
```diff
@@ -0,0 +1,165 @@
...
+<!-- vale Google.Headings = YES -->
+
+ResponseBandwidth defines a configuration to limit the speed of responding to requests.
+
+- `limit` - represented by value measure in Gbps, Mbps, kbps, or bps, for example `10kbps`
+- `percentage` - a percentage of requests on which abort will be injected, has to be in [0.0 - 100.0] range. If the value is a double number, put it in quotes.
+
+## Examples
+
+### Service backend returns 500 for 50% of requests from frontend service
```
lukidzi, line 87: Should we add better description and examples?

1. Service backend returns 500 for 50% of requests from service client in test namespace
2. 50.5% of requests to service backend from any service in namespace test (lets edit example to use a spiffeid with namespace
3. Service backend applies all faults for request from service client in namespace test 
or something simliar since the spiffeID is not showing frontend anymore

Automaat: updated examples and titles, it should make more sense now

app/_src/policies/meshfaultinjection_experimental.md
Automaat: I changed order in this section I think it should be easier to understand now

app/_src/policies/meshfaultinjection_experimental.md
Automaat: Rewritten to use cases, we don't put full example in this section in other policies. I think it makes sense to just have overview description and leave examples to its specific section

PR #2526, opened 2025-11-26
feat(tools): add tests for check version url  script

Automaat:
## Motivation

Add test coverage for check-version-urls.sh to prevent regressions

## Implementation information

Bash test script following existing pattern from test-validate-frontmatter.sh. Creates temp git repos per test to simulate real git diff scenarios. Tests cover:

- No markdown files changed
- Clean markdown with {{ page.release }}
- Hardcoded X.Y.x/X.Y.Z URLs (fail cases)
- no-version-lint comment suppression
- /assets/images/ path exclusion
- /generated/ and /raw/ directory exclusions
- .markdown extension support

## Supporting documentation

N/A


tools/test-check-version-urls.sh
```diff
@@ -0,0 +1,209 @@
...
+}
+trap cleanup EXIT
+
+# Initialize a git repo with a base commit
+setup_repo() {
+  local repo="$1"
+  mkdir -p "$repo"
+  cd "$repo"
+  git init -q -b main
+  git config user.email "test@test.com"
```
bartsmykla:
```suggestion
  git config user.email "test@kuma.io"
```

PR #2694, opened 2026-04-27
docs(otel): add MeshOpenTelemetryBackend page

bartsmykla:
## Motivation

`MeshOpenTelemetryBackend` (MOTB) lets operators define a collector once and have MeshMetric, MeshTrace, and MeshAccessLog point at it through `backendRef`. Inline `endpoint` fields on those policies are deprecated in 2.14 and will be removed in 3.0. The 2.14 docs site has no reference page for the new resource yet.

This PR adds the page.

## What changed

- `app/_src/resources/meshopentelemetrybackend.md` (new): reference doc for the resource. Migration recipe at the top, then single-collector and node-local recipes, then spec table, then `backendRef` rules, then env-var resolution, then troubleshooting with worked `DataplaneInsight` YAML for `missing`, `blocked`, and `ambiguous` signal states. Covers the env-var policy (`mode`, `precedence`, `allowSignalOverrides`), the four-layer merge order, the ambiguity rule, multi-zone gotchas (`kuma.io/display-name` label), and the mixed-version upgrade story (`feature-otel-via-kuma-dp`).
- `app/_data/docs_nav_kuma_2.14.x.yml`: nav entry under Core Resources.
- `.github/styles/config/vocabularies/Base/accept.txt`: extends `[Oo]tel` to `OTel|[Oo]tel`. `OTel` is the canonical OpenTelemetry short form per opentelemetry.io.

### DIFF
--- a/app/_src/resources/meshopentelemetrybackend.md
+++ b/app/_src/resources/meshopentelemetrybackend.md
@@ -287,7 +287,7 @@ Per signal (one block each for `traces`, `metrics`, `logs`):
 
 | Field | Description |
 |-------|-------------|
-| `enabled` | whether the signal is configured to export at all |
+| `enabled` | whether a policy targets this signal on this backend - `false` means no policy asked for it, distinct from `state: missing` (asked for but unresolved) |
 | `state` | `ready`, `blocked`, `missing`, or `ambiguous` |
 | `envAllowed` | whether `env.mode` permits environment input for this backend |
 | `envInputPresent` | whether `kuma-dp` reported any matching environment-variable keys at bootstrap |
@@ -295,7 +295,7 @@ Per signal (one block each for `traces`, `metrics`, `logs`):
 | `missingFields` | fields the merge could not produce, such as `endpoint`, `protocol`, `headers`, `client_key` |
 | `blockedReasons` | one or more of `EnvDisabledByPolicy`, `RequiredEnvMissing`, `SignalOverridesDisallowed`, `MultipleBackendsForSignal` |
 
-A signal is `ready` when the merge produces an `endpoint`. Other fields fall back to OpenTelemetry SDK defaults.
+A signal is `ready` when the merge produces an `endpoint`. Other fields fall back to OpenTelemetry SDK defaults. A signal can be `ready` and still carry `blockedReasons` - those are soft blocks (`EnvDisabledByPolicy`, `SignalOverridesDisallowed`) that tell you environment input was ignored, not that export failed. Hard blocks (`RequiredEnvMissing`, `MultipleBackendsForSignal`) move the state out of `ready`.
 
 ### Signal `missing`: no endpoint resolved
 
@@ -341,6 +341,7 @@ openTelemetry:
       enabled: true
       state: ambiguous
       envAllowed: true
+      envInputPresent: true
       blockedReasons:
       - MultipleBackendsForSignal
   - name: backend-b
@@ -348,6 +349,7 @@ openTelemetry:
       enabled: true
       state: ambiguous
       envAllowed: true
+      envInputPresent: true
       blockedReasons:
       - MultipleBackendsForSignal
 ```
@@ -362,11 +364,11 @@ Two backends resolve to the same data plane and both allow environment input. Se
 MeshOpenTelemetryBackend not found, skipping backend  name=main-collector  labels=null
 ```
 
-In multi-zone, the most common cause is a zone-authored policy referencing a Global-synced backend by `name:` instead of `labels:`. Switch to `labels: {kuma.io/display-name: <name>}`.
+In multi-zone, the most common cause is a zone-authored policy referencing a Global-synced backend by `name:` instead of `labels:`. Switch to `labels: {kuma.io/display-name: <name>}`. A backend just applied on the Global control plane can also take a few seconds to reach Zone control planes through KDS - expect short-lived `NotReferenced` and "not found" log lines while sync catches up.
 
 ### Mixed-version data planes during upgrade
 
-`backendRef` requires the data plane to advertise the `feature-otel-via-kuma-dp` feature. All 2.14 data planes do. During an upgrade where some proxies are still on 2.13, the control plane silently skips the OTel pipe route for those proxies. The signal does not export through the backend, and there is no log entry. Look at `DataplaneInsight.openTelemetry` on the affected proxies - the section will be empty for backends that depend on `backendRef`.
+`backendRef` requires the data plane to advertise the `feature-otel-via-kuma-dp` feature. All 2.14 data planes do by default. During an upgrade where some proxies are still on 2.13, the control plane silently skips the OTel pipe route for those proxies - **no log entry is emitted**. The signal does not export through the backend. Confirm by reading `DataplaneInsight.openTelemetry` on the affected proxies: no signal status entries are written for `backendRef`-based backends until the proxy advertises the feature.
 
 Inline `endpoint` configurations stay on the direct Envoy export path and keep working through the upgrade.
 
--- a/app/_src/resources/meshopentelemetrybackend.md
+++ b/app/_src/resources/meshopentelemetrybackend.md
@@ -1,0 +1,384 @@
+---
+title: MeshOpenTelemetryBackend
+description: Reference for MeshOpenTelemetryBackend, a shared OpenTelemetry collector configuration that MeshMetric, MeshTrace, and MeshAccessLog can reference instead of duplicating endpoint settings.
+keywords:
+  - MeshOpenTelemetryBackend
+  - OpenTelemetry
+  - observability
+content_type: reference
+category: resource
+---
+
+{% if_version gte:2.14.x %}
+
+<!-- vale off -->
+
+`MeshOpenTelemetryBackend` defines an OpenTelemetry collector endpoint that observability policies reference through a `backendRef`. Without it, every MeshMetric, MeshTrace, and MeshAccessLog policy carries its own copy of the collector address. With it, the address lives in one place and the policies point at it by name.
+
+Inline `endpoint` fields on those three policies still work in 2.14 but are deprecated and will be removed in 3.0. New deployments should use `backendRef`.
+
+## Migrate from inline `endpoint`
+
+1. Create one `MeshOpenTelemetryBackend` carrying the address that was inline on the policy.
+2. On each policy, replace the inline `endpoint` with `backendRef: {kind: MeshOpenTelemetryBackend, name: <backend>}`.
+3. Signal-specific fields (`refreshInterval`, `attributes`, `body`, `sampling`) stay on the policy.
+
+To move the collector later, edit the backend - the policies stay untouched.
+
+## Single collector for all three signals
+
+The most common shape: one backend resource, three policies pointing at it.
+
+{% tabs %}
+{% tab Kubernetes %}
+
+```yaml
+apiVersion: kuma.io/v1alpha1
+kind: MeshOpenTelemetryBackend
+metadata:
+  name: main-collector
+  namespace: kuma-system
+  labels:
+    kuma.io/mesh: default
+spec:
+  endpoint:
+    address: otel-collector.observability
+    port: 4317
+  protocol: grpc
+---
+apiVersion: kuma.io/v1alpha1
+kind: MeshMetric
+metadata:
+  name: all-metrics
+  namespace: kuma-system
+  labels:
+    kuma.io/mesh: default
+spec:
+  targetRef:
+    kind: Mesh
+  default:
+    backends:
+      - type: OpenTelemetry
+        openTelemetry:
+          backendRef:
+            kind: MeshOpenTelemetryBackend
+            name: main-collector
+          refreshInterval: 30s
+---
+apiVersion: kuma.io/v1alpha1
+kind: MeshTrace
+metadata:
+  name: all-traces
+  namespace: kuma-system
+  labels:
+    kuma.io/mesh: default
+spec:
+  targetRef:
+    kind: Mesh
+  default:
+    backends:
+      - type: OpenTelemetry
+        openTelemetry:
+          backendRef:
+            kind: MeshOpenTelemetryBackend
+            name: main-collector
+    sampling:
+      overall: 80
+---
+apiVersion: kuma.io/v1alpha1
+kind: MeshAccessLog
+metadata:
+  name: all-access-logs
+  namespace: kuma-system
+  labels:
+    kuma.io/mesh: default
+spec:
+  targetRef:
+    kind: Mesh
+  default:
+    backends:
+      - type: OpenTelemetry
+        openTelemetry:
+          backendRef:
+            kind: MeshOpenTelemetryBackend
+            name: main-collector
+```
+
+{% endtab %}
+{% tab Universal %}
+
+```yaml
+type: MeshOpenTelemetryBackend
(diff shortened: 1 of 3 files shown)


slonka:
Preview:

https://deploy-preview-2694--kuma.netlify.app/docs/dev/resources/meshopentelemetrybackend/

app/_src/resources/meshopentelemetrybackend.md
```diff
@@ -0,0 +1,386 @@
...
+  - MeshOpenTelemetryBackend
+  - OpenTelemetry
+  - observability
+content_type: reference
+category: resource
+---
+
+{% if_version gte:2.14.x %}
+
+<!-- vale off -->
```
slonka: This blanket disables Vale for the entire page. Consider removing this and adding any necessary terms to the vocabulary. Are there other specific terms causing Vale failures?

app/_src/resources/meshopentelemetrybackend.md
```diff
@@ -0,0 +1,386 @@
...
+# Example sidecar environment:
+#   OTEL_EXPORTER_OTLP_ENDPOINT=https://otel-gateway.observability:4318
+#   OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
+#   OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=https://tempo.observability:4318
+# Result: traces -> tempo, logs and metrics -> otel-gateway.
+```
+
+When environment input must be ignored (regulated backends), set `mode: Disabled`. When the backend is meaningless without environment input (per-tenant headers, mTLS client keys), set `mode: Required` - the signal stays `missing` until the keys are present.
+
+## Spec fields
```
slonka: Should this section (and other sections that define fields) be removed and all useful information from this section be merged with autogenerated spec renderer at the bottom? https://deploy-preview-2694--kuma.netlify.app/docs/dev/resources/meshopentelemetrybackend/#all-options

That would make openapi docs better.

Things that are here but not at the bottom:
- Platform-specific defaults (`HOST_IP` on K8s via Downward API, `127.0.0.1` elsewhere)
- Constraint info (`path` rejected when `protocol: grpc`)
- Full behavioral explanations for `env.precedence` options

app/_src/resources/meshopentelemetrybackend.md
```diff
@@ -0,0 +1,386 @@
...
+
+Every observability policy that supports OpenTelemetry has a `backendRef` field on its OTel backend block. The reference works the same way as `BackendRef` on MeshHTTPRoute: use `name` to reference a resource in the same cluster, use `labels` to reference a resource synced from another cluster.
+
+| Field | Description |
+|-------|-------------|
+| `backendRef.kind` | Must be `MeshOpenTelemetryBackend`. |
+| `backendRef.name` | `metadata.name` of the backend, for same-cluster references. |
+| `backendRef.labels` | Label selector. Required for cross-zone references because [KDS](/docs/{{ page.release }}/production/deployment/multi-zone/) appends a hash suffix to `metadata.name` on synced resources. |
+
+Exactly one of `name` or `labels` must be set. When `labels` matches more than one backend, the oldest by creation time wins.
```
slonka: just a Q: do we emit logs or warning or something when this happens?

app/_src/resources/meshopentelemetrybackend.md
```diff
@@ -0,0 +1,386 @@
...
+
+For cross-zone references, match on `kuma.io/display-name` so the resource resolves regardless of the hashed name added during sync:
```
slonka, line 228: (not strictly related to this PR) Q: are there any drawbacks on using `kuma.io/display-name` in same-cluster references?

app/_src/resources/meshopentelemetrybackend.md
```diff
@@ -0,0 +1,386 @@
...
+```yaml
+backendRef:
+  kind: MeshOpenTelemetryBackend
+  labels:
+    kuma.io/display-name: main-collector
+```
+
+### Per-zone collectors in multi-zone
+
+When zones run separate collectors, create a backend per zone and target each policy at the matching zone. Create both the backend and the policy on the Global CP - the Global CP resolves `backendRef.name` before KDS sync, so the hashed name on the zone never matters.
```
slonka: This paragraph is confusing. "create a backend per zone" reads like "create on each zone" when it means "create one for each zone, but on Global."

```suggestion
When zones run separate collectors, create one backend per zone **on the Global CP** and scope each policy to the matching zone. Because both the backend and the policy live on Global, the CP resolves `backendRef.name` before KDS sync - the hashed name that appears on zones never matters.
```

app/_src/resources/meshopentelemetrybackend.md
```diff
@@ -0,0 +1,386 @@
...
+
+If the policy is created on a zone CP and references a backend synced from Global, use `backendRef.labels` instead - the synced backend's `metadata.name` carries a hash suffix that does not match a plain `name:`.
```
slonka, line 241: more of a meta question: should we just always say to use labels?

app/_src/resources/meshopentelemetrybackend.md
```diff
@@ -0,0 +1,386 @@
...
+
+{% tabs %}
+{% tab Kubernetes %}
+
+```yaml
+apiVersion: kuma.io/v1alpha1
+kind: MeshOpenTelemetryBackend
+metadata:
+  name: main-collector
+  namespace: kuma-system
```
slonka, line 38: should we document / comment somewhere that this resource needs to be in system ns?

app/_src/resources/meshopentelemetrybackend.md
```diff
@@ -0,0 +1,386 @@
...
+            kind: MeshOpenTelemetryBackend
+            name: collector-us-east
+          refreshInterval: 30s
+```
+
+When all zones can share the same collector service name, one backend on the Global CP is enough - DNS resolves to the local collector in each zone.
+
+## Environment-variable resolution
+
+`kuma-dp` reads `OTEL_EXPORTER_OTLP_*` environment variables locally at startup and merges them with the backend config. Secret-bearing values (headers, client keys, certificates) stay local to `kuma-dp`. Bootstrap reports the present key names only, never the values.
```
slonka: "Bootstrap" is Kuma jargon that may confuse readers. Consider:

```suggestion
`kuma-dp` reads `OTEL_EXPORTER_OTLP_*` environment variables locally at startup and merges them with the backend config. Secret-bearing values (headers, client keys, certificates) stay local to `kuma-dp`. During startup, `kuma-dp` reports only which env var keys are present to the control plane, never the values.
```

app/_src/resources/meshopentelemetrybackend.md
```diff
@@ -0,0 +1,386 @@
...
+- per-signal (override shared per signal): `OTEL_EXPORTER_OTLP_TRACES_*`, `OTEL_EXPORTER_OTLP_LOGS_*`, `OTEL_EXPORTER_OTLP_METRICS_*`
+
+For each field, `kuma-dp` resolves the first available source. With default `precedence: EnvFirst`:
+
+1. signal-specific environment variable (when `allowSignalOverrides: true`)
+2. shared environment variable
+3. explicit field on the backend
+4. built-in default
+
+With `precedence: ExplicitFirst`, the explicit backend layer moves to position 1.
```
slonka: nitpick: I was a bit confused about "layer"

```suggestion
With `precedence: ExplicitFirst`, the explicit backend field moves from position 3 to 1.
```

app/_src/resources/meshopentelemetrybackend.md
```diff
@@ -0,0 +1,386 @@
...
+
+When `mode: Disabled`, environment layers are skipped entirely.
```
slonka: It's been called "source" above - let's just pick one word

```suggestion
When `mode: Disabled`, environment variables are skipped entirely (points 1 and 2).
```

app/_src/resources/meshopentelemetrybackend.md
```diff
@@ -0,0 +1,386 @@
...
+
+When `mode: Required`, missing or invalid input blocks the signal even if explicit config or defaults could otherwise fill the gap. Use `Required` when missing input should fail loud - the signal blocks, `RequiredEnvMissing` shows up in `DataplaneInsight`, and alerts fire on the silence.
```
slonka: "alerts fire on the silence" is cryptic - what does that mean? Maybe something like:

```suggestion
Use `Required` when missing input should fail loud - the signal blocks, `RequiredEnvMissing` shows up in `DataplaneInsight`, and you can alert on the absence of exported data.
```

app/_src/resources/meshopentelemetrybackend.md
```diff
@@ -0,0 +1,386 @@
...
+| `state` | `ready`, `blocked`, `missing`, or `ambiguous` |
+| `envAllowed` | whether `env.mode` permits environment input for this backend |
+| `envInputPresent` | whether `kuma-dp` reported any matching environment-variable keys at bootstrap |
+| `overrideKinds` | OTLP fields where a per-signal variable overrides the shared layer (sorted), such as `endpoint`, `protocol`, `headers`, `timeout` |
+| `missingFields` | fields the merge could not produce, such as `endpoint`, `protocol`, `headers`, `client_key` |
+| `blockedReasons` | one or more of `EnvDisabledByPolicy`, `RequiredEnvMissing`, `SignalOverridesDisallowed`, `MultipleBackendsForSignal` |
+
+A signal is `ready` when the merge produces an `endpoint`. Other fields fall back to OpenTelemetry SDK defaults. A signal can be `ready` and still carry `blockedReasons` - those are soft blocks (`EnvDisabledByPolicy`, `SignalOverridesDisallowed`) that tell you environment input was ignored, not that export failed. Hard blocks (`RequiredEnvMissing`, `MultipleBackendsForSignal`) move the state out of `ready`.
```
slonka: Not sure everyone will understand what "ready" means, maybe something like?

```
**Soft blocks** (`EnvDisabledByPolicy`, `SignalOverridesDisallowed`) mean environment input was ignored but export still works. **Hard blocks** (`RequiredEnvMissing`, `MultipleBackendsForSignal`) prevent export entirely and move the state out of `ready`.
```

app/_src/resources/meshopentelemetrybackend.md
bartsmykla: removed - added the field names to vocab, the page passes Vale clean now

app/_src/resources/meshopentelemetrybackend.md
bartsmykla: took your wording

app/_src/resources/meshopentelemetrybackend.md
bartsmykla: added a note that it must live in kuma-system on k8s

app/_src/resources/meshopentelemetrybackend.md
bartsmykla: rewrote that line with your suggestion

app/_src/resources/meshopentelemetrybackend.md
bartsmykla: switched to your wording

app/_src/resources/meshopentelemetrybackend.md
bartsmykla: applied your suggestion

app/_src/resources/meshopentelemetrybackend.md
bartsmykla: took your phrasing - 'alert on the absence of exported data' is much clearer

app/_src/resources/meshopentelemetrybackend.md
bartsmykla: good call - deleted the Spec fields table and added a one-liner pointing at All options. proto field comments already cover most of what was there (endpoint defaults, the protocol:grpc rejects path constraint). the env.mode/precedence semantics still live in the prose where they get more room than a table cell. will file a follow-up to beef up the proto comments on env.* in the kuma repo so the schema viewer can stand on its own.

app/_src/resources/meshopentelemetrybackend.md
bartsmykla: two cases. when a name does not resolve at all, the policy gets a BackendRefsResolved: False condition (reason UnresolvedBackendRefs) and the xDS plugin logs 'MeshOpenTelemetryBackend not found, skipping backend' at info level. when labels match multiple backends and the oldest-wins tie-breaker fires, nothing is emitted - that one is silent. added both to the doc.

app/_src/resources/meshopentelemetrybackend.md
bartsmykla: two real ones. first, kuma.io/display-name is only auto-set by KDS during sync - local resources don't carry it, so for same-cluster you'd have to set the label by hand. second, if two backends share the same display-name in one zone, the label match silently picks the oldest by creation time with no warning. plain name: avoids both, which is why the doc steers people to labels only for cross-zone.

app/_src/resources/meshopentelemetrybackend.md
bartsmykla: leaning no, for the reasons in the display-name thread - labels need manual setting on local resources and have a silent oldest-wins on collisions. plain name: is the simpler and safer default for same-cluster, labels are mandatory for cross-cluster because of the hash suffix. keeping the split.

app/_src/resources/meshopentelemetrybackend.md
bartsmykla: applied your suggestion verbatim. the kuma source uses the same distinction - HasHardBlockedReason in otel_pipe.go skips EnvDisabledByPolicy and SignalOverridesBlocked, so the wording matches the code.

app/_src/resources/meshopentelemetrybackend.md
```diff
@@ -0,0 +1,376 @@
...
+description: Reference for MeshOpenTelemetryBackend, a shared OpenTelemetry collector configuration that MeshMetric, MeshTrace, and MeshAccessLog can reference instead of duplicating endpoint settings.
+keywords:
+  - MeshOpenTelemetryBackend
+  - OpenTelemetry
+  - observability
+content_type: reference
+category: resource
+---
+
+{% if_version gte:2.14.x %}
```
Automaat: this is not needed as this doc is only added to 2.14 nav

app/_src/resources/meshopentelemetrybackend.md
```diff
@@ -0,0 +1,376 @@
...
+        openTelemetry:
+          backendRef:
+            kind: MeshOpenTelemetryBackend
+            name: main-collector
+```
+
+{% endtab %}
+{% tab Universal %}
+
+```yaml
```
Automaat, line 108: we miss policies referencing backend examples in universal

app/_src/resources/meshopentelemetrybackend.md
bartsmykla: good point - dropped the wrapper. the page is only registered in docs_nav_kuma_2.14.x.yml so the gate was redundant.

app/_src/resources/meshopentelemetrybackend.md
bartsmykla: added Universal-flavoured MeshMetric, MeshTrace, and MeshAccessLog examples to the Universal tab so it mirrors the Kubernetes tab.

Automaat (approved):

PR #2708, opened 2026-05-12
docs(policies): document same-namespace consumer precedence

### LINKED ISSUE #13576: Fix ordering of namespace scoped policies
### Description

Right now when me merge policies we first sort them. When sorting we take into account policy info like `role`, `origin`, `targetRef`. 

We have found an issue when multiple policies are in the same namespace, we compare them by role and ignore namespace. This leads to the situation when we have two policies in the same namespace and more broad policy targeting `Mesh` will be labelled as consumer and specific policy to MeshService or Mesh*Route will be labelled as producer, causing broader policy to override specific one.
For example:

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshTimeout
metadata:
  name: mesh-timeout-1
  namespace: kuma-demo
  labels:
    kuma.io/mesh: default
spec:
  to:
  - targetRef:
      kind: Mesh
    default:
      http:
        streamIdleTimeout: 999s
---
apiVersion: kuma.io/v1alpha1
kind: MeshTimeout
metadata:
  name: mesh-timeout-2
  namespace: kuma-demo
  labels:
    kuma.io/mesh: default
spec:
  to:
  - targetRef:
      kind: MeshService
      name: demo-app
    default:
      http:
        streamIdleTimeout: 444s
```
In this case, `mesh-timeout-1` will be labelled as `consumer` policy and `mesh-timeout-2` will be labelled as `producer` policy. Causing `mesh-timeout-1` to override `mesh-timeout-2` which does not make sense from user point of view

We need to change how we compare policies here: 
https://github.com/kumahq/kuma/blob/a8788ad310103113a61fcd7557c8954a2d4a6977/pkg/plugins/policies/core/matchers/dataplane.go#L413-L415
We need to compare policies by role only when they are from different namespaces.

This change will need to be described in UPGRADE.md as this is a breaking change on how we merge namespace scoped policies.

slonka:
## Summary
- add small note links where producer/consumer precedence is described
- document the same-namespace exception in a short section with compact tabbed examples
- make `app/_src/policies/introduction.md` pass `mise run mdlint:branch`

fix: https://github.com/kumahq/kuma/issues/13576

rendered: https://deploy-preview-2708--kuma.netlify.app/docs/2.10.x/policies/introduction/#same-namespace-exception

Be mindful of changes in https://github.com/kumahq/kuma-website/pull/2708/changes/322634ff44207b834fbfa4b7415261933788d7a7 - they were forced by linter - not all of them are related to my changes (guessing caching).


app/_src/policies/introduction.md
```diff
@@ -1161,9 +1209,54 @@ Use labels for consumer policies and name for producer policies.
...
+    - targetRef:
+        kind: Mesh
+      default:
+        connectionTimeout: 5s
+```
+
+{% endtab %}
+{% endtabs %}
+
+In this case, `consumer-default` overrides `producer-policy`.
```
Automaat: i think we should add example how to deal with this situation,  like update consumer policy, or split it into separate policies

slonka: added

Automaat (approved):

Automaat (approved):

PR #2732, opened 2026-06-02
docs(universal): add --skip-verify to dp examples

bartsmykla:
## What

Recent Kuma releases changed `kuma-dp` to verify the control plane certificate against the system trust store by default. Skipping verification is now opt-in via `--skip-verify` (`KUMA_CONTROL_PLANE_TLS_SKIP_VERIFY`); the production path is to pass the CP CA with `--ca-cert-file` / `KUMA_CONTROL_PLANE_CA_CERT_FILE`.

The universal quickstarts use a self-signed control plane, so `kuma-dp run --cp-address=https://...` now fails verification. This adds `--skip-verify` to those examples, with a warning pointing at `--ca-cert-file` for production, so the quickstarts keep working.

Reviewers: `--skip-verify` is new, so you may want to version-gate these lines for older docs.

> Changelog: skip


app/_src/quickstart/universal.md
```diff
@@ -61,9 +61,14 @@ If `kuma-cp` is running inside docker container please see [docker authenticatio
...
 For Redis:
 
+{% warning %}
+The `--skip-verify` flag turns off TLS verification of the control plane certificate and is only appropriate for this quickstart, where the control plane uses a self-signed certificate. `kuma-dp` verifies the control plane certificate by default. In production, pass the control plane CA with `--ca-cert-file=/path/to/ca.pem` (or set the `KUMA_CONTROL_PLANE_CA_CERT_FILE` environment variable) instead of skipping verification.
+{% endwarning %}
+
 ```sh
 kuma-dp run \
   --cp-address=https://localhost:5678/ \
+  --skip-verify \
```
Automaat: **blocking:** `--skip-verify` is added unconditionally, but the flag does not exist on older `kuma-dp` builds. Kuma's UPGRADE.md scopes verify-by-default and the `--skip-verify` opt-in to 2.7.26 / 2.9.16 / 2.11.14 / 2.12.11 / 2.13.7 / 2.14.x, and the e2e fix #16804 states plainly that `--skip-verify` "is not supported on older `kuma-dp` builds." This page renders for every version (`universal.md` has no version gate at all). On 2.2.x–2.6.x, 2.8.x, 2.10.x — and the patches currently pinned in `versions.yml` for the backported minors (e.g. 2.13.6 < 2.13.7) — `kuma-dp run --skip-verify` fails with `unknown flag: --skip-verify`, and the claim "`kuma-dp` verifies the control plane certificate by default" is false. Version-gate these lines. The backport is patch-granular and `{% if_version %}` is minor-granular, so exact gating is impossible at the doc layer — the practical fixes are to gate to the safe set (dev/latest where it's unambiguous) and/or bump the `versions.yml` patch pins for the backported minors so the rendered version matches reality. This is the concern you already called out.

app/_src/production/dp-config/dpp-on-universal.md
```diff
@@ -34,7 +38,8 @@ networking:
 
 kuma-dp run \
   --cp-address=https://127.0.0.1:5678 \
-  --dataplane-file=dp.yaml
+  --skip-verify \
```
Automaat: **blocking:** Same unconditional `--skip-verify` — added to a production guide with no version gate, so it breaks on every build that predates the flag. Needs the same gating as `universal.md`. The warning text here also asserts verify-by-default, which is false for those versions.

app/_src/quickstart/universal-demo.md
```diff
@@ -88,6 +84,7 @@ run kuma-dp manually, to do this run:
 ```sh
 KUMA_READINESS_PORT=9901 {% if_version gte:2.9.x %}KUMA_APPLICATION_PROBE_PROXY_PORT=9902 {% endif_version %} kuma-dp run \
   --cp-address=https://localhost:5678/ \
+  --skip-verify \
```
Automaat: **blocking:** Same unconditional `--skip-verify`, and here it lands inside the `{% if_version lte:2.9.x %}` block — explicitly rendered for 2.9.x and older, where the flag is unsupported on the pinned patch. Gate to the versions that actually ship the flag.

app/_src/quickstart/universal-docker-demo.md
```diff
@@ -396,10 +396,15 @@ This section explains how to start the `kv` service, which mimics key/value stor
...
 
+      {% warning %}
+      The `--skip-verify` flag turns off TLS verification of the control plane certificate and is only appropriate for this quickstart, where the control plane uses a self-signed certificate. `kuma-dp` verifies the control plane certificate by default. In production, pass the control plane CA with `--ca-cert-file=/path/to/ca.pem` (or set the `KUMA_CONTROL_PLANE_CA_CERT_FILE` environment variable) instead of skipping verification.
+      {% endwarning %}
+
       ```sh
       runuser --user {{ kuma-data-plane-proxy }} -- \
         /usr/local/bin/kuma-dp run \
           --cp-address https://control-plane:5678 \
+          --skip-verify \
```
Automaat: **blocking:** Same unconditional `--skip-verify` in the Docker quickstart. Gate to the supported versions; on older `kuma-dp` images this fails with `unknown flag`.

app/_src/quickstart/universal-docker-demo-meshidentity.md
```diff
@@ -408,10 +408,15 @@ This section explains how to start the `kv` service, which mimics a key/value st
...
 
+      {% warning %}
+      The `--skip-verify` flag turns off TLS verification of the control plane certificate and is only appropriate for this quickstart, where the control plane uses a self-signed certificate. `kuma-dp` verifies the control plane certificate by default. In production, pass the control plane CA with `--ca-cert-file=/path/to/ca.pem` (or set the `KUMA_CONTROL_PLANE_CA_CERT_FILE` environment variable) instead of skipping verification.
+      {% endwarning %}
+
       ```sh
       runuser --user {{ kuma-data-plane-proxy }} -- \
         /usr/local/bin/kuma-dp run \
           --cp-address https://control-plane:5678 \
+          --skip-verify \
```
Automaat: **blocking:** Same unconditional `--skip-verify` in the MeshIdentity Docker quickstart. Note MeshIdentity is itself recent, so the safe floor here is narrower than the other files — confirm the floor version.

app/_src/production/dp-config/dpp-on-universal.md
```diff
@@ -19,6 +19,10 @@ To avoid users bypassing the sidecar, have the service listen only on the intern
 
 For example, this is how we start a `Dataplane` for a hypothetical Redis service and then start the `kuma-dp` process:
 
+{% warning %}
+The `--skip-verify` flag turns off TLS verification of the control plane certificate and is only appropriate for this guide, where the control plane uses a self-signed certificate. `kuma-dp` verifies the control plane certificate by default. In production, pass the control plane CA with `--ca-cert-file=/path/to/ca.pem` (or set the `KUMA_CONTROL_PLANE_CA_CERT_FILE` environment variable) instead of skipping verification.
```
Automaat: **issue:** `dpp-on-universal.md` is a production how-to (`content_type: how-to`, under `production/`), not a quickstart. Its primary example now leads with `--skip-verify`, and readers copy the command, not the warning above it. A production guide's default example should use `--ca-cert-file=/path/to/ca.pem`; if you want to show the self-signed shortcut, make it the secondary/aside, not the first thing the reader pastes.

app/_src/production/dp-config/dpp-on-universal.md
```diff
@@ -34,7 +38,8 @@ networking:
 
 kuma-dp run \
```
Automaat, line 39: **praise:** The PR fixes a pre-existing broken line continuation here — the original `--dataplane-file=dp.yaml` lacked a trailing backslash, orphaning `--dataplane-token-file` on the next line so it was never passed to `kuma-dp run`. Adding the `\` repairs the command.

app/_src/quickstart/universal.md
Automaat: **Applied** (52f1ab07) — gated `--skip-verify` and the warning to `gte:2.11.x`, and bumped the 2.11/2.12/2.13 pins in `versions.yml` to their latest patches (2.11.15 / 2.12.12 / 2.13.8) so the rendered version actually ships the flag. With that, the flag renders only where verify-by-default exists; 2.8.x/2.10.x got no backport so they stay excluded, and 2.7.x/2.9.x stay pinned pre-flag (2.7.25 / 2.9.15) so they render without it — internally consistent at each pinned patch.

One correction to the premise: `universal.md` is the quickstart only for 2.2.x–2.5.x (2.6–2.8 use `universal-demo.md`, 2.9+ `universal-docker-demo.md`), so the `gte:2.11.x` gate resolves to empty here and the flag no longer renders on any version this page serves.

app/_src/production/dp-config/dpp-on-universal.md
Automaat: **Applied** (52f1ab07) — this is a production how-to, so the pasted command no longer uses `--skip-verify`. It now passes `--ca-cert-file=/path/to/ca.pem`, which works on every version and verifies by default, so no version gate is needed on the command itself. See the thread on line 23 for the full rework.

app/_src/quickstart/universal-demo.md
Automaat: **Applied** (52f1ab07) — removed `--skip-verify` from the `lte:2.9.x` command variant and gated the `gte:2.10.x` variant to `gte:2.11.x`. `universal-demo.md` only serves 2.6.x–2.8.x (all pre-flag at their pinned patch), so the flag no longer renders on this page at all.

app/_src/quickstart/universal-docker-demo.md
Automaat: **Applied** (52f1ab07) — all three `kuma-dp run --skip-verify` lines plus the warning are gated to `gte:2.11.x`. Verified against a local build: 2.9.x/2.10.x render the command without the flag (continuation intact) and 2.11.x+ render it with the flag. The `kumactl … --skip-verify` on line 242 is a different, pre-existing flag and left untouched.

app/_src/quickstart/universal-docker-demo-meshidentity.md
Automaat: **Applied** (52f1ab07) — gated to `gte:2.13.x`. Confirmed the floor: the MeshIdentity quickstart is nav-featured only from 2.13.x (`docs_nav_kuma_2.13.x` / `2.14.x`), so that's the narrowest safe floor, and both 2.13.8 and dev ship the flag.

app/_src/production/dp-config/dpp-on-universal.md
Automaat: **Applied** (52f1ab07) — reworked so the default pasted command uses `--ca-cert-file=/path/to/ca.pem` (the production-correct path; version-agnostic). `--skip-verify` is now only a secondary aside in the warning, gated to `gte:2.11.x`, not something the reader pastes.

app/_src/production/dp-config/dpp-on-universal.md
Automaat: Thanks — the backslash continuation is preserved. The command now passes `--ca-cert-file=/path/to/ca.pem` and then `--dataplane-file=dp.yaml \` with the continuation intact (52f1ab07).

Automaat:
Merged `master` and resolved the `versions.yml` conflict (7d6b8989).

Heads up — master changed the gating math, so I widened the version gates beyond the earlier `gte:2.11.x`:

- master already bumped the pins (took master's `versions.yml`): now **2.7.27**, **2.9.17**, **2.11.16**, **2.12.12**, **2.13.8**, plus released **2.14.0** (latest) and **2.15.x** (dev).
- That means 2.7.x and 2.9.x now render a patch that ships `--skip-verify` (verify-on-by-default), so they need the flag too. The backported set is now `{2.7, 2.9, 2.11, 2.12, 2.13, 2.14, 2.15}` with gaps at 2.8 and 2.10 (no backport).

Gates per page (each maps to the quickstart that version actually serves):

- `universal.md` (2.2–2.5) → no flag versions; gate never renders.
- `universal-demo.md` (2.6–2.8) → `eq:2.7.x`.
- `universal-docker-demo.md` (2.9–2.15) → `eq:2.9.x` + `gte:2.11.x` (excludes 2.10).
- `universal-docker-demo-meshidentity.md` (2.13–2.15) → `gte:2.13.x`.
- `dpp-on-universal.md` (production, all versions) → command uses `--ca-cert-file`; `--skip-verify` only a `gte:2.11.x` aside.

Verified against a full build: `--skip-verify` renders on 2.7 / 2.9 / 2.11–2.15 and is absent (continuation intact) on 2.2–2.6 / 2.8 / 2.10. No new markdownlint/vale errors vs the pre-merge tree.

Automaat (approved):

PR #2748, opened 2026-06-23
fix(nav): use dev nav when version nav missing

### LINKED ISSUE #1987: Prevent navigation to new version leading to 404
### What happened?

After #1958 got merged, the dropdown navigation showed the new version `2.9.x`. After clicking on it, the 404 page showed up.

Screenshot 2024-10-17 at 12 54 19

### How to prevent this in the future?

Would be great to add a check that for each version in the [`versions.yaml`](https://github.com/kumahq/kuma-website/blob/master/app/_data/versions.yml) there is a respective `docs_nav_kuma_<version>.yaml` in [app/_data](https://github.com/kumahq/kuma-website/tree/master/app/_data) to prevent this in the future. Consider extending the [kuma workflow](https://github.com/kumahq/kuma/blob/master/.github/workflows/update-docs.yaml) accordingly.

Automaat:
## Motivation

A version added to `app/_data/versions.yml` without its own `docs_nav_kuma_<version>.yml` had no pages generated for it: the `jekyll-generator-single-source` gem builds docs pages by globbing `docs_nav_*.yml`, so a missing nav file means a missing page set. The version dropdown lists every `versions.yml` entry regardless, so selecting the new version led to a 404 (originally seen after the 2.9.x sync added the version without its nav file).

> Changelog: fix(nav): fall back to dev nav for versions missing their own nav

## Implementation information

The `Versions` generator now, before tagging pages, fills the gap for any non-dev version that lacks its own nav: it exposes the dev nav under that version (so the sidebar renders) and generates the version's pages from the dev nav config retargeted at the release, so URLs and `{% if_version %}` gating resolve to the correct version.

A guard skips versions whose generated raw assets (crds/protos/rbac) were not synced yet, keeping the old behaviour (no pages) instead of breaking the whole build on assets those pages would load. In practice the update-docs job syncs a version's raw assets together with its `versions.yml` entry (the 2.9.x sync added the full raw tree), so the common case — only the nav lagging — is covered.

Verified with a real build: removing an existing version's nav file (assets present) regenerates its full page set from dev (136 pages, matching dev), with sidebar links scoped to that version. New specs cover the fallback, the asset guard, the own-nav case, and the no-dev case.

## Supporting documentation

Closes #1987.


app/_plugins/generators/versions.rb
```diff
@@ -38,6 +43,58 @@ def generate(site)
...
+
+        # The update-docs job syncs a version's raw assets (crds/protos/rbac)
+        # together with its versions.yml entry, so they are normally present when
+        # only the nav lags. If a version was added far ahead of its assets,
+        # skip it: keep the old behaviour (no pages) instead of breaking the
+        # whole build on the missing assets its pages would load.
+        next unless version_assets?(site, release)
+
+        # Expose the dev nav under this version so the sidebar renders.
+        site.data[key] = dev_nav
```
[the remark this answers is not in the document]
Automaat: Done in d2596dbd — the dev nav is now exposed retargeted: `site.data[key] = dev_nav.merge('release' => release)`, so the nav copy carries the version's release and is no longer the shared dev hash.

app/_plugins/generators/versions.rb
```diff
@@ -38,6 +43,58 @@ def generate(site)
...
+
+        # Generate the version's pages from the dev nav config, retargeted at
+        # this release so URLs and version gating resolve to the right version.
+        nav_config = generator_config.build_docs_nav(edition: dev['edition'], release: dev['release'])
+        nav_config.config['release'] = release
+        nav_config.generate_pages
```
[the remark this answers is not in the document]
Automaat: The suggested alternative is not feasible here: `build_docs_nav(release: release)` builds the path `docs_nav_kuma_<version>.yml` and reads it eagerly (`NavConfig::Base.process` → `File.read`), but that file is exactly what is missing, so it would raise `ENOENT`. Sourcing dev's file then retargeting `config['release']` is necessary. On the safety of the mutation: in gem 0.0.15 `DocNavConfig#initialize` only stores `i18n_file`/`site` — no path/state is precomputed from release; `release`, `items` and `versions` are all lazy (`@x ||=`) and resolved during `generate_pages`, after the mutation. `build_docs_nav` also returns a fresh config each call (`File.read` per call), so there is no cross-iteration state. Added a code comment explaining this.

app/_plugins/generators/versions.rb
```diff
@@ -38,6 +43,58 @@ def generate(site)
...
+      return unless dev_nav
+
+      generator_config = Jekyll::GeneratorSingleSource::GeneratorConfig.new(site)
+
+      site.data['versions'].each do |version|
+        next if version['label'] # labeled releases (e.g. dev) are generated by the gem
+
+        release = version['release']
+        key = nav_key(release)
+        next if site.data[key] # version has its own nav already
```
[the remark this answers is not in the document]
Automaat: Done in d2596dbd — `GeneratorConfig` is now a lazily memoized accessor (`@generator_config ||= ...`, matching the existing `edition` helper in this file), so it is only built once a version actually needs fallback generation.

spec/generators/versions_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    generator.send(:generate_missing_version_navs, site)
+  end
+
+  context 'when a non-dev version has no nav file of its own' do
+    let(:versions) { [{ 'edition' => 'kuma', 'release' => '2.13.x' }, dev] }
+
+    it 'exposes the dev nav under the missing version so the sidebar renders' do
+      run!
+
+      expect(data['docs_nav_kuma_213x']).to be(dev_nav)
```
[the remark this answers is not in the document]
Automaat: Done in d2596dbd — the spec now asserts on structure (`fallback['items'] == dev_nav['items']`, `fallback['release'] == '2.13.x'`) plus `not_to be(dev_nav)`, matching the new copy-and-retarget behaviour rather than object identity.

app/_plugins/generators/versions.rb
```diff
@@ -38,6 +43,67 @@ def generate(site)
...
+    # docs_nav_kuma_<version>.yml, generate pages and a sidebar from the dev nav.
+    def generate_missing_version_navs(site)
+      dev = site.data['versions'].detect { |v| v['label'] == 'dev' }
+      return unless dev
+
+      dev_nav = site.data[nav_key(dev['release'])]
+      return unless dev_nav
+
+      site.data['versions'].each do |version|
+        next if version['label'] # labeled releases (e.g. dev) are generated by the gem
```
slonka: `next if version['label']` skips every labeled version, not just `'dev'`. If another entry ever gets one (a new tag, a future label), it gets silently excluded from this fallback


```suggestion
        next if version['label'] == 'dev' # labeled releases (e.g. dev) are generated by the gem
```

app/_plugins/generators/versions.rb
```diff
@@ -38,6 +43,67 @@ def generate(site)
...
+    def generate_version_from_dev(site, dev, dev_nav, release)
+      # Expose the dev nav, retargeted at this version, so the sidebar renders
+      # and the nav's release matches the version rather than dev.
+      site.data[nav_key(release)] = dev_nav.merge('release' => release)
+
+      # The version's own docs_nav file is absent, so load the dev nav config and
+      # retarget it at this release: pages then resolve URLs and version gating to
+      # the right version. build_docs_nav(release: release) is not an option here,
+      # it would try to read the missing docs_nav_kuma_<version>.yml.
+      nav_config = generator_config(site).build_docs_nav(edition: dev['edition'], release: dev['release'])
```
slonka, line 85: Two related things here:

1. `build_docs_nav(edition:, release: dev['release'])` is called once per missing version with constant args, so the dev `docs_nav_kuma_<dev>.yml` is parsed N times.
2. The reason it has to be in the loop is the next-line mutation `nav_config.config['release'] = release`, which only works because `DocNavConfig#release` / `#items` are lazily memoized (mutate first, then `generate_pages` resolves through `config['release']`).

That's a load-bearing assumption about a gem internal. A short comment pinning it ("DocNavConfig#release / #items are lazy, so mutate before generate_pages") would protect a future reader from either inlining the YAML load or tripping over a gem change. The N reads are likely cheap, just calling it out.

spec/generators/versions_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+  let(:site) { instance_double(Jekyll::Site, data: data, config: config) }
+
+  let(:generator_config) { instance_double(Jekyll::GeneratorSingleSource::GeneratorConfig) }
+  let(:nav_config) { instance_double(Jekyll::GeneratorSingleSource::DocNavConfig, config: {}, generate_pages: nil) }
+
+  before do
+    allow(Jekyll::GeneratorSingleSource::GeneratorConfig).to receive(:new).with(site).and_return(generator_config)
+    allow(generator_config).to receive(:build_docs_nav).and_return(nav_config)
+    # By default every release has its raw assets present.
+    allow(Dir).to receive(:exist?).and_return(true)
```
slonka: Global `allow(Dir).to receive(:exist?).and_return(true)` stubs every `Dir.exist?` call for the whole example. RSpec / Jekyll / gem internals can call `Dir.exist?` and silently get `true`. Narrowing to the asset path it's meant to cover (`with(File.join('app/assets', release, 'raw'))`) keeps the suite robust against incidental callers.

spec/generators/versions_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+    end
+
+    it 'generates the version pages from the dev nav retargeted at the version' do
+      run!
+
+      expect(generator_config).to have_received(:build_docs_nav).with(edition: 'kuma', release: '2.15.x')
+      expect(nav_config.config['release']).to eq('2.13.x')
+      expect(nav_config).to have_received(:generate_pages)
+    end
+  end
```
slonka, line 60: Consider a spec for two missing versions in one run, each ending up with its own retargeted release. The `nav_config.config['release'] = release` pattern is correct today because `build_docs_nav` returns a fresh `DocNavConfig` per call, but a test would pin that invariant and catch a regression if the gem ever started memoizing the instance (the current contexts only exercise one missing version each).

app/_plugins/generators/versions.rb
Automaat: Applied in b4516252 — `next if version['label'] == 'dev'`. Note: the fallback is release-keyed (nav_key, asset path, retargeted `config['release']` all use the release value), and `version_assets?` checks `app/assets/<release>/raw`. A labeled version's assets live under `<label>/raw`, so such an entry is still skipped by the asset guard rather than mis-generated — but narrowing to `== 'dev'` is the correct intent and avoids silently excluding a future non-dev label.

app/_plugins/generators/versions.rb
Automaat: Done in b4516252 — expanded the comment to pin it: "Load-bearing gem invariant: DocNavConfig#release and #items are lazily memoized, so we must mutate config['release'] BEFORE generate_pages, and a fresh DocNavConfig is needed per version (so the dev file is parsed once per missing version - cheap, but worth noting)." Left the N parses as-is per your call.

spec/generators/versions_spec.rb
Automaat: Done in b4516252 — the blanket stub is now `allow(Dir).to receive(:exist?).and_call_original`, and each context stubs only its own path via a `stub_assets(release, present)` helper (`with(File.join('app/assets', release, 'raw'))`).

spec/generators/versions_spec.rb
Automaat: Added in b4516252 — new context "with several versions missing their nav in one run" (2.12.x + 2.13.x). `build_docs_nav` returns a fresh `DocNavConfig` double per call and captures `config['release']` at generate_pages time; the test asserts each ends up retargeted independently (`contain_exactly('2.12.x', '2.13.x')`), pinning the per-call freshness invariant.

slonka (approved):

